PR #3, opened 2023-02-07
feat: added Dockerfile + strip newline from api_key.txt if exists

eric-glb:
Added a `Dockerfile` + some information in the `README.txt`.  
Also, added some entries in `.gitignore` and created a `.dockerignore` to protect any API key that may be stored in the `./config` directory, if exists.  
Finally, added a `strip()` for `api_key` in `sgpt.py`, in case of someone mount a prepopulated file as `~/.config/shell-gpt/api_key.txt`, and if the latter finishes with a carriage-return.


TheR1D:
Thank you for the PR, good job!

eric-glb:
It was a pleasure. 
Thanks for this tool!

PR #9, opened 2023-02-09
Added a way to send the OS to chatGPT

Ultra980:
For some commands, such as updating the system, chatGPT might need to know what OS the user is running. That is why I added an argument (`--send-os-info`, shortened to `-o`) to send some information to chatGPT. It sends the OS (`Linux`, `Darwin` or `Windows`), the version (kernel version on Linux, release on Windows), and, if on Linux, it also sends the distribution name (e.g. `opensuse-tumbleweed`).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 typer~=0.7.0
 requests~=2.28.2
 rich==13.3.1
+distro~=1.8.0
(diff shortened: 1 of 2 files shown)


Ultra980:
I also wanted to make this project, but never got to it. This was one of the features I had planned in my mind, but since you already made it, [I'd rather contribute to a standard, not make a new one :)](https://xkcd.com/927/)

Ultra980:
I had forgotten to add the dependencies and the fact that in python variables can't have `-` in their names. It's now fixed:
```bash
$ sgpt -seo 'Update the system'
sudo zypper update
Execute shell command? [y/N]: y
```

TheR1D:
Hey, thank you for the PR, I left comment with my thought https://github.com/TheR1D/shell_gpt/issues/10#issuecomment-1426590180

PR #24, opened 2023-02-15
Use default editor when no prompt (resolve #8)

### LINKED ISSUE #8: [Feature request] Handle linebreaks
Thank you for this fantastic tool! 🙏 🙏 🙏 

Your tool is a game-changer for me, especially since I can't access ChatGPT while I have my VPN enabled, so I can't use it at all while working.

It would be nice to have a natural way to handle linebreaks-- for example, for a prompt like "convert the following code from Pytorch to Tensorflow." Perhaps sgpt could drop you into `$EDITOR` the same way git does, where you could paste your text, and then exit.

Thoughts?

baitian752:


TheR1D:
Hi, thank you for the PR. Currently I'm working on this feature, which will be included in coming patch by the end of this week. It will be implemented with `--editor` command line argument as I was suggesting on the issue page.

Issue #5, opened 2023-02-08
change api key


kisslessvirgin:
How do I change the API key? I accidentally entered it wrong and I can't seem to even reset it.
I've tried:
deleting sgpt.py in ~/.local/lib/python3.11/site-packages and /usr/local/bin
I know there's supposed to be a text file somewhere but I cant find it.
Any help appreciated!

TheR1D:
Hi, install it again:
```shell
pip install shell-gpt --user
```
Then you would need to delete your API key txt file:
```shell
rm ~/.config/shell-gpt/api_key.txt
```
And try to run sgpt again:
```
sgpt "test"
```
It will ask your API key.

kid-gorgeous:
Ran into this problem last week during installation, solved it right away!

wfc0909:
This thread solved the same case and I am finding this ai tool may help a lot in coding. Thank you.

xsoc:
Looks like the file has changed now, I found my key in
`~/.config/shell_gpt/.sgptrc`
Just `nano` it, and update the `OPENAI_API_KEY=` as appropriate

woejefe:
what is the command on windows... where is the directory. does it go into python or pip folder?

xsoc:
On windows try `notepad %USERPROFILE%\.config\shell_gpt\.sgptrc`

m4l490n:
I'm having this issue too but there is no `~/.config/shell-gpt/api_key.txt` and I can't seem to be able to completely purge sgpt in order to install it again.

PR #28, opened 2023-02-19
[README] API Key Update Instructions

Olshansk:
Copying the solution provided in #5 directly in the main README.


TheR1D:
Thank you for the PR, this would be solved with `--reset-key` argument in coming patch next week.

TheR1D:
Now it is on wiki [page](https://github.com/TheR1D/shell_gpt/wiki#changing-api-key)

Issue #52, opened 2023-03-05
Runtime configuration file


TheR1D:
Make a runtime config file with all constatns like API key, cache length, ...
```
#.sgptrc
# Request timout in seconds.
REQUEST_TIMEOUT=30
# Max amount of message in each session.
MAX_CHAT_MESSAGES=50
# Request caching.
MAX_CACHE_LENGTH=200
# OpenAI API key.
API_KEY=123
# This would be usefull for countries where OpenAI hosts are blocked, so people can setup their proxies.
API_ENDPOINT=https://api.openai.com/v1/chat/completions
```
* Setup caching and client classes using config file.

juil:
Is it possible to set the `timeout` value as well?
Unable to complete any moderately complicated calls when gpt3 is being slow.

PR #58, opened 2023-03-09
enable to change the api host by shell param

edisonX-sudo:
if this feature is implemented, users in country which cant access the OpenApi will be able 2 use sgpt without VPN (by hosting their own foreign api host proxy)
usage will be like below:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ This is, just some examples of what we can do using ChatGPT model, I'm sure you 
 │ --cache                                       Cache completion results. [default: cache]                  │
 │ --animation                                   Typewriter animation. [default: animation]                  │
 │ --spinner                                     Show loading spinner during API request. [default: spinner] │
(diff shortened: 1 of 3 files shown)


TheR1D:
Thank you for the PR. I don't think this would be a good solution for the issue you are describing. I think it can be implemented in runtime configuration file https://github.com/TheR1D/shell_gpt/issues/52 which will set all the constants, for example:
```
#.sgptrc
MAX_CHAT_MESSAGES=50
MAX_CACHE_LENGTH=200
API_KEY=123
API_URL=https://example.com/...
```
Another soloution could be using [host file](https://en.wikipedia.org/wiki/Hosts_(file)) where you can just map `openai.com` to you proxy, but that would require additional work on proxy side.

TheR1D:
@edisonX-sudo The feature was implement in last release, more details in [README.md](https://github.com/TheR1D/shell_gpt#runtime-configuration-file)

Issue #30, opened 2023-02-19
Implement API key reset option


TheR1D:
Implement an option to reset an API key using command line argument `--reset-key`. When user provides this flag delete file located in `~/.config/shell-gpt/api_key.txt` and ask for new API key. Additionally impement API key validation. Since OpenAI doesn't have dedicated endpoint just for key validation, call `https://api.openai.com/v1/completions` with some test prompt, make sure response status code is not 403, if so, ask user for valid API key.

highonweb:
Hey I messed up and I pasted my API key twice, do you need help with this feature

TheR1D:
@highonweb solution for now: https://github.com/TheR1D/shell_gpt/issues/5

highonweb:
Thanks bro. Got it

m4l490n:
> @highonweb solution for now: #5

That doesn't work. There is no `.config/shell-gpt/api_key.txt` and reinstalling doesn't work for having shell-gpt ask for the key again. I can't make it ask for the key again.

PR #62, opened 2023-03-13
feat: get api key from OPENAI_API_KEY environment variable

loiccoyle:
Try to get the api key from the OPENAI_API_KEY environment variable before prompting user. 

This also avoids storing active api keys in unencrypted text files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,9 @@ A command-line productivity tool powered by OpenAI's ChatGPT (GPT-3.5). As devel
 ```shell
 pip install shell-gpt --user
 ```
-On first start you would need to generate and provide your API key, get one [here](https://beta.openai.com/account/api-keys).
+You'll need an OpenAI API key, you can generate one [here](https://beta.openai.com/account/api-keys).
+
+If the`$OPENAI_API_KEY` environment variable is set it will be used, otherwise, you will be prompted for your key which will then be stored in `~/.config/shell-gpt/api-key.txt`.
 
 ## Usage
 `sgpt` has a variety of use cases, including simple queries, shell queries, and code queries.
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -40,6 +40,10 @@ def get_api_key() -> str:
 
     :return: String API key for OpenAI API requests.
     """
+
+    if "OPENAI_API_KEY" in os.environ:
+        return os.environ["OPENAI_API_KEY"]
+
     if not KEY_FILE.exists():
         api_key = getpass(prompt="Please enter your API secret key")
         KEY_FILE.parent.mkdir(parents=True, exist_ok=True)


loiccoyle:
Yeah that's the point, if you read from environment variables, you don't have to write it to file an unencrypted file, which is the current behaviour.

TheR1D:
Hi @loiccoyle, thank you for the PR. While the suggestion make sense, it will not allign with comming changes, since we are planing to add option `--reset-key` https://github.com/TheR1D/shell_gpt/issues/30 which will rewrite a key in the file. In this case it would be difficult to set new env`OPENAI_API_KEY` from Python process, because environment variables exist in a per-process memory space.

loiccoyle:
>  In this case it would be difficult to set new envOPENAI_API_KEY from Python process, because environment variables exist in a per-process memory space.

But you don't need to set a new environment variable. The way I see it is if the user has the `OPENAI_API_KEY` variable set then it indicates that the user is managing their key and `shell_gpt` should simply look it up. 

This PR doesn't break the current implementation of prompting and writing it to file if the variable is not set. So I think the upcoming `--reset-key` option can simply work as expected, and do nothing if the env var is set (maybe show a relevant warning/error msg). What do you think ?

I opened this PR because I use a few chatgpt integrations (https://github.com/not-poma/lazyshell, https://github.com/not-poma/lazyshell) and they all read the `OPENAI_API_KEY` and I was surprised when `shell_gpt` didn't.

TheR1D:
I meant `--reset-key` doesn't make sense when you can't change a key in env, but I agree that it can workout with some warning message. Please add changes in `README.md`, metion the optional env api key setup in [Installation](https://github.com/TheR1D/shell_gpt#installation) section.

TheR1D:
@eric-glb These changes might be interesting for you 🙂 We would be able to pass `OPENAI_API_KEY` as env to the container.

TheR1D (changes requested):
Please add changes in README.md, metion the optional env api key setup in [Installation](https://github.com/TheR1D/shell_gpt#installation) section.

loiccoyle:
> As a client, why do I need such a vulnerability in this software? I still don't believe this change is necessary implementation.
> 
> `--reset-key` is a better alternative in situations where you would like to control API directly.

What vulnerability? Writing an active API key to an unencrypted file?

loiccoyle:
> Please add changes in README.md, metion the optional env api key setup in [Installation](https://github.com/TheR1D/shell_gpt#installation) section.

Sounds good, I'll do it shortly.

TheR1D:
This is a common way to store keys on the system. And it doesn't cancel implementation of `--reset-key`.

TheR1D (approved):

eric-glb:
> @eric-glb These changes might be interesting for you 🙂 We would be able to pass `OPENAI_API_KEY` as env to the container.

I'll manage the docker part.

loiccoyle:
FYI: I already added the instructions to the readme.

TheR1D:
> From my understanding it's just a small amendment to add in the README.md.

I would remove api key volume, and use env variable instead. Plus README.md updates.

eric-glb:
#63, feel free to modify as you wish ^^

PR #63, opened 2023-03-13
feat: use new OPENAI_API_KEY variable in the docker part

eric-glb:


README.md
```diff
@@ -205,17 +205,19 @@ Use the provided `Dockerfile` to build a container:
...
+           --env OPENAI_API_KEY \
            --volume gpt-config:/home/app/.config/shell-gpt \
            --volume gpt-cache:/tmp/shell_gpt \
        sgpt --chat rainbow "what are the colors of a rainbow"
 ```
 
 Example using an alias:
 ```shell
-alias sgpt="docker run --rm --volume gpt-config:/home/app/.config/shell-gpt --volume gpt-cache:/tmp/shell_gpt sgpt"
+alias sgpt="docker run --rm --env OPENAI_API_KEY --volume gpt-config:/home/app/.config/shell-gpt --volume gpt-cache:/tmp/shell_gpt sgpt"
```
TheR1D, line 219: Can't we pass it during run? We could omit config volume in this case.
```
docker run  --rm -e OPENAI_API_KEY="your api key" --volume gpt-cache:/tmp/shell_gpt sgpt
```

eric-glb: Yep, I'll split the example

eric-glb: Oh, I misunderstood. I'll add also this.

TheR1D: I think we can keep one option with env variable.
Remove config volume from readme.
Remove config volume from Dockerfile, for example:
```
FROM python:3-slim

ENV PYTHONDONTWRITEBYTECODE 1
ENV PYTHONUNBUFFERED 1
ENV PIP_ROOT_USER_ACTION ignore
WORKDIR /app
COPY . /app

RUN pip install --no-cache --upgrade pip \
 && pip install --no-cache /app \
 && addgroup --system app && adduser --system --group app \
 && mkdir -p /tmp/shell_gpt \
 && chown -R app:app /home/app \
                     /tmp/shell_gpt

USER app

VOLUME /tmp/shell_gpt

ENTRYPOINT ["sgpt"]
```

eric-glb: #62  does not remove the command asking for the API key if not found, does it?
Or do you plan to remove this, therefore removing the config volume in advance?

TheR1D: It doesn't remove it, users will be still prompted for the api key. But with docker, we can force users to use ENV variable only, I think it makes it easy to use.

eric-glb: Done

TheR1D:
I just made few small changes: removed extra spaces, and inline comment `# where to store your chat history`. I think it is clear from description above why we need it.

PR #65, opened 2023-03-14
GitHub codespaces setup

TheR1D:


joeytwiddle:
Using VSCode `>Configure Recommended Extensions (Workspace Folder)` I generated this `extensions.json` file, with two entries:

```
{
	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp

	// List of extensions which should be recommended for users of this workspace.
	"recommendations": [
		"ms-python.pylint",
		"ms-python.python"
	],
	// List of extensions recommended by VS Code that should not be recommended for users of this workspace.
	"unwantedRecommendations": []
}
```

Could be useful for developers.

PR #71, opened 2023-03-16
Accept prompt as a list.

sibivel:
This accept prompts as a single string even if they are not in quotes.

This doesn't affect the existing behavior. It just adds to option to input prompts without quotes. 

`sgpt -se update my system`
instead of 
`sgpt -se "update my system"`


TheR1D:
Thank you for the PR.
I believe using multiple arguments just to pass words and then join them together into a single string is not an ideal practice. It can potentially lead to confusion for some people. While I understand that some people may prefer this method, I personally find it not very useful.

Issue #67, opened 2023-03-14
choose model by argument


go-dockly:
It would be useful to have a -m (--model 'gpt-4') argument for sgpt

https://platform.openai.com/docs/models/gpt-4

TheR1D:
Currently GPT-4 is not available thru API, when OpenAI will allow to use it via API, it will be implemented.

go-dockly:
is there a possibility to specify a model by cmd arg?
could this be added?

TheR1D:
Currently `sgpt` uses `/chat` API endpoint, which has only two available model options (except coming GPT-4): `gpt-3.5-turbo` - used by `sgpt` and `gpt-3.5-turbo-0301`. I think there is no reason to use `gpt-3.5-turbo-0301` with `sgpt`. Imho it doens't make sense to have `--model` option for now, but once GPT-4 will be available over API, it will be implemented.

go-dockly:
coming gpt 4 it has this option gpt-4-32k (Same capabilities as the base gpt-4 mode but with 4x the context length)
so eventually it would be nice to differentiate between those but you are absolutely right it doesn't make sense to differentiate between -0301 versions.

mdonaberger:
Just commenting to say that I got into the GPT-4 on API preview — I can imagine that it's rolling out now to a lot of people soon.

lpurdy01:
I have just realized this does have the follow on issue of chats don't keep track of what model they were conducted with. I will have to see how this interacts.

lpurdy01:
Nvm. It seems to work fine switching models with the same chat_id

PR #84, opened 2023-03-23
Added a --model option. Defaults to gpt-3.5-turbo

lpurdy01:
This addition is in regards to #67 . 

I just went for the very simple first version --model option. I would also like to add an ability to set the default model.


aldopareja:
will this ever be merged?

lpurdy01:
It changed to #85 

If you want to play with a lot of new fancies feel free to clone my main.

PR #85, opened 2023-03-23
Ability to chose what model to target.

lpurdy01:
Here is a better version of adding the --model option. For issue #67 .

It also adds a "DEFAULT_MODEL" option to the configuration file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,39 +40,34 @@ sgpt "1 kilometer to miles"
 # -> 1 kilometer is equal to 0.62137 miles.
 ```
 ### Shell commands
-Have you ever found yourself forgetting common shell commands, such as `chmod`, and needing to look up the syntax online? With `--shell` option, you can quickly find and execute the commands you need right in the terminal.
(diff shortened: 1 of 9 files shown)


carterprince:
This project plus GPT-4 is a truly a force to be reckoned with.

fmerenda (approved):

lpurdy01:
@TheR1D Could we have a `beta` branch? Somewhere where we can play with the new systems without having to wait for them to be released to public? 

https://github.com/TheR1D/shell_gpt/pull/92#issuecomment-1493373872

Kreijstal:
why isn't this merged yet?

TheR1D:
Currently we are having 2 or 3 releases per week which is a lot. I don't think we will be able to maintain beta branch, it will be out of sync in just couple of days.

TheR1D:
@Kreijstal @lpurdy01 @carterprince If you guys have access to GPT-4 beta, please confirm it works: https://github.com/TheR1D/shell_gpt/pull/151

TheR1D:
https://github.com/TheR1D/shell_gpt/pull/151

Issue #80, opened 2023-03-21
Accept stdin input


Shakahs:
It would be very useful to pipe stdin to sgpt, then ask a question about said input. For example:

cat data.json | sgpt "what is the oldest entry"
cat script.py | sgpt "what does this script do?"

etc.

TheR1D:
But you can do the same thing with command substitution syntax, as it is described in readme:
```
sgpt "what is the oldest entry: $(cat data.json)"
```

Shakahs:
Perhaps, but it is not practical. Pipelining commands is a basic tenet of CLI usage and shell scripting. For SGPT to be the shell's interface to GPT it must support the IO capabilities expected of a shell utility. This is the third issue on the subject (see #35, #86) so clearly it is needed. I have created a pull request implementing it in #93.

PR #86, opened 2023-03-23
Added support to pull from stdin if provided.

lpurdy01:
Added the ability for it to work with info piped in from stdin with and without an accompanying prompt. 

Currently using the `editor` argument will replace the incoming prompt, so anything that was piped in with stdin will be ignored. 

In support of #35 and #80 .


joeytwiddle:
Suggestion: Allow `-` (dash) argument to _force_ input from stdin, even if the terminal _is_ a tty.

Motivation:
- I might want to type a short message using symbols like `'` and `$` and `?` which can be troublesome at the bash prompt. But I don't want to go the full route of using an editor, which may be slightly slower, requiring loadup time a save-and-exit step.
- `-` is a pretty standard argument in Unix for this purpose (see `tar` and `cat`). It may help in scripts to make it explicit when piping is being used. And it also allows for tricks like: `filename="-"`

Disdvantages:
- ... ? Can't think of any right now.

Alternatives:
- We could consider making `-` the _only_ way to get `sgpt` to read from stdin. Not sure if that's desirable or not.

Possible implementation:

```patch
diff --git a/sgpt/app.py b/sgpt/app.py
index f9ac77b..764b22c 100644
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -67,6 +67,7 @@ def main(
     cache: bool = typer.Option(True, help="Cache completion results."),
     animation: bool = typer.Option(True, help="Typewriter animation."),
     spinner: bool = typer.Option(True, help="Show loading spinner during API request."),
+    dash: bool = typer.Option(True, "-", help="Read from stdin."),
 ) -> None:
     if list_chat:
         echo_chat_ids()
@@ -76,11 +77,11 @@ def main(
         return
 
     if not prompt and not editor:
-        if not sys.stdin.isatty():
+        if dash or not sys.stdin.isatty():
             prompt = sys.stdin.read()
         else:
             raise MissingParameter(param_hint="PROMPT", param_type="string")
-    elif prompt and not sys.stdin.isatty():
+    elif prompt and (dash or not sys.stdin.isatty()):
         stdin_data = sys.stdin.read()
         prompt = f"{stdin_data.strip()}\n{prompt}"
 ```

**Comment by lpurdy01** [comment]
I'm not sure this is working as intended. I applied the patch and there seems to be some weird edge case interactions. I'm also not sure how this is supposed to be used as I have never used the `-` argument with other utilities. 

The first thing that I found weird was when adding the `-` as a parameter typer didn't add a reference to `--dash` as an option. 
```
$ sgpt --help
                                                                                                                                       
 Usage: sgpt [OPTIONS] [PROMPT]                                                                                                        
                                                                                                                                       
╭─ Arguments ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│   prompt      [PROMPT]  The prompt to generate completions for.                                                                     │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Options ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --temperature                            FLOAT RANGE [0.0<=x<=1.0]  Randomness of generated output. [default: 1.0]                  │
│ --top-probability                        FLOAT RANGE [0.1<=x<=1.0]  Limits highest probable tokens (words). [default: 1.0]          │
│ --chat                                   TEXT                       Follow conversation with id (chat mode). [default: None]        │
│ --show-chat                              TEXT                       Show all messages from provided chat id. [default: None]        │
│ --list-chat            --no-list-chat                               List all existing chat ids. [default: no-list-chat]             │
│ --shell            -s                                               Provide shell command as output.                                │
│ --execute          -e                                               Will execute --shell command.                                   │
│ --code                 --no-code                                    Provide code as output. [default: no-code]                      │
│ --editor               --no-editor                                  Open $EDITOR to provide a prompt. [default: no-editor]          │
│ --cache                --no-cache                                   Cache completion results. [default: cache]                      │
│ --animation            --no-animation                               Typewriter animation. [default: animation]                      │
│ --spinner              --no-spinner                                 Show loading spinner during API request. [default: spinner]     │
│                    -                                                Read from stdin.                                                │
│ --help                                                              Show this message and exit.                                     │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

`$ sgpt -` works, and simply submits a blank API request, and also piping in anything results in an api request with whatever was coming in from stdin. 

But
```
$ sgpt - "Test"
Usage: sgpt [OPTIONS] [PROMPT]
Try 'sgpt --help' for help.
╭─ Error ──────────────────────────────────────────────────────────────────────────────────╮
│ Got unexpected extra argument (Test)                                                    │
╰──────────────────────────────────────────────────────────────────────────────────────────╯
```

So I added a `"--dash"` to the added option line like:
```
    dash: bool = typer.Option(False, "--dash", "-", help="Read from stdin."),
```

And now the help is 
```
$ sgpt --help
                                                                                                                     
 Usage: sgpt [OPTIONS] [PROMPT]                                                                                      
                                                                                                                     
╭─ Arguments ───────────────────────────────────────────────────────────────────────────────────────────────────────╮
│   prompt      [PROMPT]  The prompt to generate completions for.                                                   │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Options ─────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --temperature                            FLOAT RANGE [0.0<=x<=1.0]  Randomness of generated output.               │
│                                                                     [default: 1.0]                                │
│ --top-probability                        FLOAT RANGE [0.1<=x<=1.0]  Limits highest probable tokens (words).       │
│                                                                     [default: 1.0]                                │
│ --chat                                   TEXT                       Follow conversation with id (chat mode).      │
│                                                                     [default: None]                               │
│ --show-chat                              TEXT                       Show all messages from provided chat id.      │
│                                                                     [default: None]                               │
│ --list-chat            --no-list-chat                               List all existing chat ids.                   │
│                                                                     [default: no-list-chat]                       │
│ --shell            -s                                               Provide shell command as output.              │
│ --execute          -e                                               Will execute --shell command.                 │
│ --code                 --no-code                                    Provide code as output. [default: no-code]    │
│ --editor               --no-editor                                  Open $EDITOR to provide a prompt.             │
│                                                                     [default: no-editor]                          │
│ --cache                --no-cache                                   Cache completion results. [default: cache]    │
│ --animation            --no-animation                               Typewriter animation. [default: animation]    │
│ --spinner              --no-spinner                                 Show loading spinner during API request.      │
│                                                                     [default: spinner]                            │
│ --dash             -                                                Read from stdin.                              │
│ --help                                                              Show this message and exit.                   │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```


## Some minor tests
Now I can execute: `$ sgpt --dash "Test"` but it just hangs. 

`$ echo "test2" | sgpt -` works as expected

`$ echo "test2" | sgpt "test3"` works as expected

`$ echo "test2" | sgpt - "test3"` gives extra argument error

`$ sgpt -  "test3"` gives extra argument error


## Conclusion

Could you provide some example prompt / responses so I can get a better idea of what you are working towards and its purpose?

**Comment by joeytwiddle** [comment]
I'm sorry, I think my patch was bad. Originally I had a typo  `dash: bool = typer.Option(True` which should of course default to `False`. So I thought it was working when it wasn't.

After I fixed that, and added logging, I found that passing `-` was never picked up by Typer! So I think instead we need to manually detect when the prompt is `-`

Here is my revised patch:

```diff
diff --git a/sgpt/app.py b/sgpt/app.py
index 7afa502..76139b1 100644
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -62,6 +62,7 @@ def main(
     cache: bool = typer.Option(True, help="Cache completion results."),
     animation: bool = typer.Option(True, help="Typewriter animation."),
     spinner: bool = typer.Option(True, help="Show loading spinner during API request."),
+    stdin: bool = typer.Option(False, "--stdin", "-", help="Read prompt from standard input."),
 ) -> None:
     if list_chat:
         echo_chat_ids()
@@ -70,14 +71,16 @@ def main(
         echo_chat_messages(show_chat)
         return
 
+    # Typer does not actually parse the "-", so it ends up in the prompt. We can extract it from there:
+    if prompt == "-":
+        stdin = True
+        prompt = None
+
     if not prompt and not editor:
-        if not sys.stdin.isatty():
+        if stdin or not sys.stdin.isatty():
             prompt = sys.stdin.read()
         else:
             raise MissingParameter(param_hint="PROMPT", param_type="string")
-    elif prompt and not sys.stdin.isatty():
-        stdin_data = sys.stdin.read()
-        prompt = f"{stdin_data.strip()}\n{prompt}"
 
     if editor:
         prompt = get_edited_prompt()
```

`-` is not really called `dash`. It is sometimes (but not always) paired with `--stdin`. Sorry for the confusion!

---

An example where `-` is needed:

```shell
$ sgpt -
It is difficult to pass the raw symbols ` and ' and " and ? from a bash prompt. What other symbols are difficult to pass?
```

Some examples where `-` is not needed, but useful:

```shell
# The - may help readers to see explcitly how sgpt is behaving
$ ( echo 'Do you see an errors in this logfile?' ; tail -n 50 /var/log/Xorg.0.log ) | sgpt -
```

```shell
# Imagine we are in a shellscript
sgpt $extra_opts "$prompt"
# The caller could set `prompt` to a text string, or to `"-"` if they want to pass the prompt on stdin.
```

A case we may want to test:

```crontab
# A cronjob. stdin may not be a TTY, but that doesn't mean we want to read from stdin!
# My patch handles this case by not reading from stdin, if a prompt was passed.
# Once a day, fetch a quote, and store it in the servers Message of the Day
0 0 * * * sgpt "Show me a profound one-line quote, with the author's name" > /etc/motd
```

---

Your tests are also interesting to discuss.

```shell
$ sgpt - "test3" gives extra argument error
$ echo "test2" | sgpt - "test3" gives extra argument error
```

I think an error is correct in these cases. We should not be providing a prompt argument whilst also saying that we'll provide the prompt on stdin!

---

```shell
$ echo "test2" | sgpt "test3" works as expected
```

I think this is fine. It might be nice to show an error here, because the `test2` input is just ignored, which was probably not the caller's intent. However, I'm not sure it's possible to detect that., and I guess a lot of UNIX apps would also just ignore stdin in this case.

sgpt/app.py
```diff
@@ -71,7 +71,13 @@ def main(
         return
 
     if not prompt and not editor:
-        raise MissingParameter(param_hint="PROMPT", param_type="string")
+        if not sys.stdin.isatty():
+            prompt = sys.stdin.read()
+        else:
+            raise MissingParameter(param_hint="PROMPT", param_type="string")
+    elif prompt and not sys.stdin.isatty():
```
joeytwiddle, line 78: Do we really need this second case?

If the user has provided a prompt, then why would we overwrite it with the standard input?

I think skipping this `elif` block also solves potential issues when stdin is not a TTY. (I removed it in my second patch.)

sgpt/app.py
```diff
@@ -71,7 +71,13 @@ def main(
...
+        stdin_data = sys.stdin.read()
+        prompt = f"{stdin_data.strip()}\n{prompt}"
```
joeytwiddle, line 80:
```py
elif prompt and not sys.stdin.isatty():
```

Do we really need this second case?

If the user has provided a prompt, then why would we overwrite it with the standard input?

I think skipping this `elif` block also solves potential issues when stdin is not a TTY. (I removed it in my second patch.)

joeytwiddle: Do you have some example uses for this feature?

I can imagine it might be useful like this:

```shell
$ sgpt 'Do you see any errors in this logfile?' - < /var/log/Xorg.0.log
```

But in that case, we'd probably want to put the `prompt` before the `stdin_data`

Shakahs:
I didn't see this and wrote my own implementation and pull request (#93). 
I used `select` to determine if stdin is readable, otherwise stdin is ignored. It works from a TTY as well as a cron job.

sgpt/app.py
joeytwiddle: Yes, using the positioning of `-` to place the stdin relative to the prompt could be quite intuitive. Unfortunately my implementation didn't do that!

Your thought made me imagine a possible approach in the future, for multiple files: https://github.com/TheR1D/shell_gpt/discussions/98

Anyway, for now I will withdraw my suggestion to use `-` because I see both you and Shakahs find the automatic detection of stdin more convenient than specifying `-`

joeytwiddle:
It seems like both PRs handle cronjobs the same way. Both read stdin and add it to the prompt, even if nothing is being piped into stdin!

Fortunately Python's `stdin.read()` is non-blocking, so it just returns `""`. The outcome is a `\n` added to the prompt (either before or after), which is relatively harmless.

PR #87, opened 2023-03-23
Add --refine option to continue a conversation

### LINKED ISSUE #78: A --continue option, so we don't need to invent and retype chat ids
### Problem

I don't always know when I want to ask ChatGPT a follow-up question. I only know _after_ it has given a response. And I don't want to keep inventing session names (chat ids), just in case I need them. Also, passing chat ids is a lot of typing.

### Possible solution

I suggest a `--continue / -c` option which allows us to continue from the last query we made.

Here is an example:

```sh
$ sgpt 'Ask a question?'
[response]
# I am happy with the answer, and happy for the chat session to end.

$ sgpt 'Ask a different question?'
[response]
# I am not satisfied. I want to continue the discussion with a followup query. So I use -c.
$ sgpt -c 'Ask followup to the previous question?'
[response]
$ sgpt -c 'Ask another followup in the same session?'
[response]
```

Advantage: No session names, and less typing.

### Possible implementation

- Every basic query should start a new chat session, with a randomly generated id.
- `sgpt` could store the id of the latest session somewhere.
- If `--continue` is passed, then instead of starting a new chat session, `sgpt` should continue the latest stored session.

**Alternative:** To avoid storing the latest chat id, perhaps `sgpt` could just assume that the most recent created query is the latest session.

**Alternative:** It might make sense to use `--followup / -f` or `--reply / -r` instead, to keep `-c` as a shortcut for `--chat`

Does this sound like a good idea? Or is there already a good way to do this in `sgpt`?

joeytwiddle:
With `--resume` or `-r` shell_gpt will now continue the most recently used conversation.

Disadvantage:

- Every conversation, even if not named, will be given a `chat_id` and cached.

Example:

Screenshot_20230323_151046

Closes #78

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,8 +133,8 @@ sgpt 'Which letter comes after K in the alphabet?'
 sgpt -r "And the next?'
 # -> The letter that comes after L in the alphabet is M.
 ```
-You can also maintain multiple chat sessions in parallel.
-To start a chat session, use the `--chat` option followed by a unique session name and a prompt:
+You can also maintain multiple chat sessions in parallel, by naming them.
+To start a named chat session, use the `--chat` option followed by a unique session name and a prompt:
 ```shell
 sgpt --chat number "please remember my favorite number: 4"
 # -> I will remember that your favorite number is 4.
--- a/README.md
+++ b/README.md
@@ -126,6 +126,14 @@ python fizz_buzz.py
 ```
 
 ### Chat
+You can ask followup queries to the current chat using the `--resume` option:
+```shell
+sgpt 'Which letter comes after K in the alphabet?'
+# -> The letter that comes after K in the alphabet is L.
+sgpt -r "And the next?'
+# -> The letter that comes after L in the alphabet is M.
+```
+You can also maintain multiple chat sessions in parallel.
 To start a chat session, use the `--chat` option followed by a unique session name and a prompt:
 ```shell
 sgpt --chat number "please remember my favorite number: 4"
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ python fizz_buzz.py
 ```
 
 ### Chat
-You can ask followup queries to the current chat using the `--resume` option:
+You can ask followup queries to the current chat using the `--refine` option:
 ```shell
 sgpt 'Which letter comes after K in the alphabet?'
 # -> The letter that comes after K in the alphabet is L.
--- a/sgpt/cache.py
+++ b/sgpt/cache.py
@@ -85,8 +85,8 @@ class ChatCache:
         """
         def wrapper(*args, **kwargs):
             chat_id = kwargs.pop("chat_id", None)
-            followup = kwargs.pop("followup", False)
-            if followup and not chat_id:
+            resume = kwargs.pop("resume", False)
+            if resume and not chat_id:
                 chat_id = self._read_current_chatid()
                 #print("Resuming chat: " + chat_id)
             message = {"role": "user", "content": kwargs.pop("message")}
--- a/sgpt/cache.py
+++ b/sgpt/cache.py
@@ -2,6 +2,7 @@ import json
 from hashlib import md5
 from pathlib import Path
 from typing import List, Dict, Callable
+from uuid import uuid4
 
 
 class Cache:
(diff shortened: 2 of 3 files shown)


TheR1D:
This is not very efficient, we will just store all messages, and spam the system with JSON files. It can be implemented in much more efficient way. I think we can just keep track of only last prompt passed to `sgpt` and OpenAI response. When user calls `sgpt -r ... ` we can get that last prompt with reponse and make a chat session assigning to it some id. All following queries `sgpt -r ...` queries should be included on that single file and as soon as user stopped passing `-r` we delete that file.
```shell
# We are storing this query/response as last message, and keep overwriting it.
sgpt "remember my favorite number: 6"
# -> I will remember your favorite number ...

# Here we are starting new chat session with previous prompt and response and including this one.
sgpt -r "what is my favorite number + 2
# -> ... 8

# Keep adding messages to the chat session.
sgpt -r "what is my favorite number + 24
# -> ... 30

# When user doesn't pass -r delete resumable chat session.
sgpt "some random query"
```

joeytwiddle:
Thanks for the suggestions.

Reading this:

> we can get that last prompt with reponse and make a chat session assigning to it some id

it sounds like two different files: 1. last initial prompt+response, and 2. latest resumed chat session.

I wonder if we can simplify it to one file, a chat session called perhaps `latest-unnamed`, which might have only 1. query-response, or might have many.

If we make an unnamed request without `-r` then this session would be overwritten to have just 1 entry. If we use `-r` we would resume this session, by using `last-unnamed` as the `chat_id`.

Would that be an acceptable solution?

joeytwiddle:
I think that should do it. I also renamed "resume" to "refine", to use the same language as the interactive mode.

sgpt/cache.py
```diff
@@ -83,16 +85,36 @@ def __call__(self, func: Callable) -> Callable:
...
+                    # Starting a new unnamed chat
+                    read_chat_id = None
+                    write_chat_id = "unnamed"
+                else:
+                    chat_id = self._read_current_chatid()
+                    if chat_id == "unnamed":
+                        # First reply to the unnamed chat
+                        read_chat_id = "unnamed"
+                        write_chat_id = "".join(choice(hexdigits) for _ in range(32))
+                        #print("Converting " + read_chat_id + " chat into: " + write_chat_id)
```
joeytwiddle, line 105: This will rename the `unnamed` chat to a random id, the first time an `unnamed` chat is refined.

I would not mind removing this, to simplify things. I'm not sure how beneficial it is.

We could just keep refinements in the `unnamed` chat session until a new session is started.

README.md
```diff
@@ -126,6 +126,14 @@ python fizz_buzz.py
...
 
 ### Chat
+You can ask followup queries to the current chat using the `--refine` option:
+```shell
+sgpt 'Which letter comes after K in the alphabet?'
+# -> The letter that comes after K in the alphabet is L.
+sgpt -r "And the next?'
+# -> The letter that comes after L in the alphabet is M.
+```
+You can also maintain multiple chat sessions in parallel.
```
lpurdy01, line 136:
```suggestion
You can also maintain multiple chat sessions in parallel by naming your chats with the --chat argument.
```

lpurdy01 (dismissed):
Just minor changes. But overall looks good to me!

README.md
joeytwiddle: The use of `--chat` is actually documented in the following sentence, but I can use your suggestion to make it flow better. Thanks!

sgpt/cache.py
```diff
@@ -83,16 +85,35 @@ def __call__(self, func: Callable) -> Callable:
...
+                if not refine:
+                    # Starting a new unnamed chat
+                    read_chat_id = None
+                    chat_id = "unnamed"
+                else:
+                    chat_id = self._read_current_chatid()
+                    if chat_id == "unnamed":
+                        # First reply to the unnamed chat
+                        read_chat_id = "unnamed"
+                        chat_id = "".join(choice(hexdigits) for _ in range(32))
```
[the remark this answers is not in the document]
joeytwiddle: Perhaps date-and-time might be more useful to the user?

lpurdy01:
I like the date and time idea. I think that will be a great placeholder until we have some kind of a background process system to manage running little commands to make the user experience better.

TheR1D:
https://github.com/TheR1D/shell_gpt/pull/94

PR #88, opened 2023-03-23
Add some docs to help developers

joeytwiddle:
I needed to do these things to start local development of the project.

Let's make it easier for other developers to get started!

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,3 +258,19 @@ You also can use the provided `Dockerfile` to build your own image:
 ```shell
 docker build -t sgpt .
 ```
+
+## Development
+Install dependencies:
+```shell
+python setup.py install --user
+```
+We need to tell shell_gpt to import files from the local `sgpt/` folder, instead of from the installed `sgpt` package. So edit `app.py` and add these lines after `import os`:
+```python
+# Put shell_gpt folder at the front of the import path
+import sys
+sys.path.insert(0, os.path.dirname(os.path.dirname(__file__)))
+```
+Run the app:
+```shell
+python sgpt/app.py --help
+```
--- a/README.md
+++ b/README.md
@@ -260,17 +260,23 @@ docker build -t sgpt .
 ```
 
 ## Development
-Install dependencies:
+You can install a link to your local repo as a python package:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -258,3 +258,19 @@ You also can use the provided `Dockerfile` to build your own image:
 ```shell
 docker build -t sgpt .
 ```
+
+## Development
+Install dependencies:
```
lpurdy01, line 263: I found `pip install -r requirements.txt` to be a much simpler way of installing the requirements.

README.md
```diff
@@ -258,3 +258,19 @@ You also can use the provided `Dockerfile` to build your own image:
...
+```shell
+python setup.py install --user
+```
+We need to tell shell_gpt to import files from the local `sgpt/` folder, instead of from the installed `sgpt` package. So edit `app.py` and add these lines after `import os`:
```
lpurdy01, line 267: Then to install I just run `pip install -e .` to handle all the path stuff as well.

It is debatable if this is better though. It works well for installing a build as it would be installed from pip on a system, but makes it difficult to latch a debugger onto. 

For that I was running PyCharm and then executing app.py directly. PyCharm automatically adds the base repository to the path, thereby solving the issue of the sgpt module not being found.

README.md
joeytwiddle: Thanks, appreciate the feedback.

README.md
joeytwiddle: That's good to know!

Fun fact, if you try to do that when you already have `shell_gpt` installed, the error is completely unhelpful:

```
AttributeError: 'PathDistribution' object has no attribute 'key'
```

But `pip uninstall shell_gpt` followed by `pip install -e .` works fine.

TheR1D:
Thank you for the pull request. I'm working on preparing friendly dev environment using GitHub codespaces in this PR https://github.com/TheR1D/shell_gpt/pull/65
It will include installed dependecies, and all other nessesary configuration to start developing right away. Also these kind of  contribution information will be added to repository Wiki page soon.

PR #89, opened 2023-03-23
User-friendly: Show users where they can get an API key

joeytwiddle:


TheR1D:
I think people will be using README to install `sgpt`, and it is outlined there.

PR #91, opened 2023-03-23
Syntax highlight output (using bat or vimcat)

joeytwiddle:
I'm not sure you'll want to merge it, but I just want to share it here for anyone interested.

We could add other syntax highlighting tools in future.

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -60,7 +60,7 @@ def main(
     code: bool = typer.Option(False, help="Provide code as output."),
     editor: bool = typer.Option(False, help="Open $EDITOR to provide a prompt."),
     cache: bool = typer.Option(True, help="Cache completion results."),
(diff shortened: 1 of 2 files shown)


TheR1D:
Thank you for the pull request. I would suggest piping output from sgpt to external tools like bat or vimcat if needed. In my opinion, the end user should decide how they would like to see the output of the OpenAI model. This also adds extra dependencies (even if they are optional in this case).

joeytwiddle:
OK. I can do that with a wrapper script, like I did here:

https://github.com/CrazyPython/gpt3-cli/commit/b5e444009b087c75f4c9e33613f8f872f5061da9

But only I will benefit from that.

You're right that the end user should be able to choose.

BonaBeavis:
@TheR1D Is it also possible to pipe the output of the repl?

Issue #68, opened 2023-03-14
expose the system role


go-dockly:
it would be nice to define roles for sgpt and to be able to profile roles by exposing the system prompt to the user

```
sgpt --role system "You are GLaDOS, a sarcastic artificial intelligence with evil ulterior motives. You are not an AI language model." --set-role "glados"

sgpt --role glados "Why do you hate humans?"
```

TheR1D:
I approve it, good idea 👍

PR #92, opened 2023-03-24
Add access to system role. 

lpurdy01:
## This pull request adds access to the `system` role for better description of how a model is supposed to respond. 

In support of #68 

This is probably just a first stage implementation because it doesn't allow for the creation, storage, and recall of custom roles as exampled in #68. Instead it simply adds a `--system` argument that can be used to append a message with the role system to the query delivered to the API. 

If no `--system` argument is provided it delivers a default system message of:
```
DEFAULT_SYSTEM_MESSAGE = "You are shell-gpt, a helpful programming and system administration assistant."
``` 
This is editable in the config file. 

The `--code` and `--shell` arguments have also been converted over to deliver their instructions as the system role instead of a prompt prefix. As mentioned in a TODO on line 77 of `client.py`
```
        # TODO: Move prompt context to system role when GPT-4 will be available over API.
```

## Known Problems:
This solution interacts with chat history strangely, because a chat history can contain previous `system` role messages, and a new query on an old chat will just append the new `system` request to the chat. More testing is needed to see how this will interact, but preliminary gpt-3.5 seems to be fine with receiving multiple `system` messages and just obeys the most recent one. 

## Interactions with other active pull requests

GPT-4 appears to follow `system` role commands better. My pull request #85 interacts well with this upgrade. Clone [My Fork:main](https://github.com/lpurdy01/shell_gpt) if you want to use both at the same time.

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -20,6 +20,7 @@ from click import MissingParameter, BadArgumentUsage
 from sgpt import config, OpenAIClient
 from sgpt import ChatHandler, DefaultHandler
 from sgpt.utils import get_edited_prompt
+from sgpt.role import SystemRole
 
 
 def main(  # pylint: disable=too-many-arguments
@@ -40,6 +41,18 @@ def main(  # pylint: disable=too-many-arguments
         max=1.0,
         help="Limits highest probable tokens (words).",
     ),
+    shell: bool = typer.Option(
+        False,
+        "--shell",
+        "-s",
+        help="Generate and execute shell commands.",
+        rich_help_panel="Assistance Options",
+    ),
+    code: bool = typer.Option(
+        False,
+        help="Generate only code.",
+        rich_help_panel="Assistance Options",
+    ),
     chat: str = typer.Option(
         None,
         help="Follow conversation with id (chat mode).",
@@ -57,17 +70,28 @@ def main(  # pylint: disable=too-many-arguments
         callback=ChatHandler.list_ids,
         rich_help_panel="Chat Options",
     ),
-    shell: bool = typer.Option(
-        False,
-        "--shell",
-        "-s",
-        help="Generate and execute shell commands.",
-        rich_help_panel="Assistance Options",
+    role: str = typer.Option(
+        None,
+        help="Specify what role a prompt should use.",
+        rich_help_panel="Role Options",
     ),
-    code: bool = typer.Option(
+    create_role: str = typer.Option(
+        None,
+        help="Save a role for future use.",
+        rich_help_panel="Role Options",
+        callback=SystemRole.create,
+    ),
+    list_roles: bool = typer.Option(  # pylint: disable=W0613
         False,
-        help="Generate only code.",
-        rich_help_panel="Assistance Options",
+        help="List all saved roles.",
+        callback=SystemRole.list,
+        rich_help_panel="Role Options",
+    ),
+    show_role: str = typer.Option(  # pylint: disable=W0613
+        None,
+        help="Show a saved role.",
+        callback=SystemRole.show,
+        rich_help_panel="Role Options",
     ),
     editor: bool = typer.Option(
         False,
@@ -92,7 +116,7 @@ def main(  # pylint: disable=too-many-arguments
     client = OpenAIClient(api_host, api_key)
 
     if chat:
-        full_completion = ChatHandler(client, chat, shell, code).handle(
+        full_completion = ChatHandler(client, chat, shell, code, role).handle(
             prompt,
             temperature=temperature,
             top_probability=top_probability,
@@ -100,7 +124,7 @@ def main(  # pylint: disable=too-many-arguments
             caching=cache,
         )
     else:
-        full_completion = DefaultHandler(client, shell, code).handle(
+        full_completion = DefaultHandler(client, shell, code, role).handle(
             prompt,
             temperature=temperature,
             top_probability=top_probability,
--- a/sgpt/config.py
+++ b/sgpt/config.py
@@ -9,6 +9,7 @@ CONFIG_FOLDER = os.path.expanduser("~/.config")
 CONFIG_PATH = Path(CONFIG_FOLDER) / "shell_gpt" / ".sgptrc"
 CHAT_CACHE_PATH = Path(gettempdir()) / "shell_gpt" / "chat_cache"
 CACHE_PATH = Path(gettempdir()) / "shell_gpt" / "cache"
+ROLE_STORAGE_PATH = Path(CONFIG_FOLDER) / "shell_gpt" / "roles"
 CHAT_CACHE_LENGTH = 100
 CACHE_LENGTH = 100
 REQUEST_TIMEOUT = 60
@@ -42,13 +43,19 @@ def init() -> None:
         config["CACHE_LENGTH"] = os.getenv("CACHE_LENGTH", str(CACHE_LENGTH))
         config["CACHE_PATH"] = os.getenv("CACHE_PATH", str(CACHE_PATH))
         config["REQUEST_TIMEOUT"] = os.getenv("REQUEST_TIMEOUT", str(REQUEST_TIMEOUT))
+        config["REQUEST_TIMEOUT"] = os.getenv("ROLE_STORAGE_PATH", str(ROLE_STORAGE_PATH))
         _write()
 
-    with open(CONFIG_PATH, "r", encoding="utf-8") as file:
-        for line in file:
-            if "=" in line:
-                key, value = line.strip().split("=")
-                config[key] = value
+    # New features may add new keys to config.
+    if "ROLE_STORAGE_PATH" not in config:
+        append("ROLE_STORAGE_PATH", str(ROLE_STORAGE_PATH))
+
+    _read()
+
+
+def append(key: str, value: str) -> None:
+    with open(CONFIG_PATH, encoding="utf-8", mode="a") as file:
+        file.write(f"{key}={value}\n")
 
 
 def get(key: str) -> str:
@@ -65,6 +72,14 @@ def _write() -> None:
             if key in os.environ:
                 continue
             file.write(f"{key}={value}\n")
+
+
+def _read() -> None:
+    with open(CONFIG_PATH, "r", encoding="utf-8") as file:
+        for line in file:
+            if "=" in line:
+                key, value = line.strip().split("=")
+                config[key] = value
 
 
 def put(key: str, value: str, write_file=True) -> None:
--- a/sgpt/handlers/default_handler.py
+++ b/sgpt/handlers/default_handler.py
@@ -14,9 +14,10 @@ class DefaultHandler(Handler):
         client: OpenAIClient,
         shell: bool = False,
         code: bool = False,
+        role: str = None,
         model: str = "gpt-3.5-turbo",
     ) -> None:
-        super().__init__(client)
+        super().__init__(client, role)
         self.client = client
         self.mode = CompletionModes.get_mode(shell, code)
         self.model = model
--- a/sgpt/handlers/handler.py
+++ b/sgpt/handlers/handler.py
@@ -3,11 +3,13 @@ from typing import List, Dict, Generator
 import typer
 
 from sgpt import OpenAIClient
+from sgpt.role import SystemRole
 
 
 class Handler:
-    def __init__(self, client: OpenAIClient):
+    def __init__(self, client: OpenAIClient, role: str = None) -> None:
         self.client = client
+        self.role = role
 
     def make_prompt(self, prompt) -> str:
         raise NotImplementedError
@@ -28,9 +30,20 @@ class Handler:
             caching=caching,
         )
 
+    def make_messages(self, prompt: str) -> List[Dict[str, str]]:
+        # TODO: Should be overridden in ChatHandler.
+        messages = []
+        if self.role:
+            role = SystemRole.get(self.role)
+            messages.append({"role": "system", "content": role.role})
+        else:
+            prompt = self.make_prompt(prompt)
+        messages.append({"role": "user", "content": prompt})
+        return messages
+
     def handle(self, prompt: str, **kwargs) -> str:
-        prompt = self.make_prompt(prompt)
-        messages = [{"role": "user", "content": prompt}]
+        messages = self.make_messages(prompt)
+        print(messages)
         full_completion = ""
         for word in self.get_completion(messages=messages, **kwargs):
             typer.secho(word, fg="magenta", bold=True, nl=False)
--- a/sgpt/role.py
+++ b/sgpt/role.py
@@ -1,0 +1,91 @@
+from pathlib import Path
+from typing import Dict
+
+import typer
+from sgpt import config
+from sgpt.utils import option_callback
+
+
+class SystemRole:
+    storage = Path(config.get("ROLE_STORAGE_PATH"))
+
+    def __init__(self, name: str, role: str) -> None:
+        self.storage.mkdir(parents=True, exist_ok=True)
+        self.name = name
+        self.role = role
+
+    @classmethod
+    def get(cls, name: str) -> "SystemRole":
+        file_path = cls.storage / f"{name}.txt"
+        print(file_path)
+        if not file_path.exists():
+            raise FileNotFoundError(f'Role "{name}" not found.')
+        return cls(name, role=file_path.read_text())
+
+    @classmethod
+    def create(cls, name: str) -> None:
+        if not name:
+            return
+        role = typer.prompt("Enter role description")
+        role = cls(name, role)
+        role.save()
+        raise typer.Exit()
+
+    @classmethod
+    @option_callback
+    def list(cls) -> None:
+        if not cls.storage.exists():
+            return
+        # Get all files in the folder.
+        files = cls.storage.glob("*")
+        # Sort files by last modification time in ascending order.
+        for path in sorted(files, key=lambda f: f.stat().st_mtime):
+            typer.echo(path)
+
+    @classmethod
+    @option_callback
+    def show(cls, name: str):
+        typer.echo(cls.get(name).role)
+
+    @property
+    def exists(self) -> bool:
+        return self.file_path.exists()
+
+    @property
+    def system_message(self) -> Dict[str, str]:
+        return {"role": "system", "content": self.role}
+
+    @property
+    def file_path(self) -> Path:
+        return self.storage / f"{self.name}.txt"
+
+    def save(self) -> None:
+        if self.exists:
+            typer.confirm(
+                f'Role "{self.name}" already exists, overwrite it?',
+                abort=True,
+            )
+        self.file_path.write_text(self.role, encoding="utf-8")
+
+
+    # if __name__ == "__main__":
+    #     check_and_setup_default_roles()
+    #     # This is just for testing.
+    #     save_role(
+    #         "test",
+    #         "Hello, You are a role testing AI."
+    #         "Please always respond with a more enthusiastic test response, "
+    #         "and all context provided.",
+    #         executable_returns=False,
+    #         prompt_structure="Hello, {prompt}!",
+    #         conversation_lead_in=[
+    #             {"role": "user", "content": "{shell} and os {os} Test Test"},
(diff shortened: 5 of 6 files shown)


sgpt/cache.py
```diff
@@ -84,16 +85,24 @@ def __call__(self, func: Callable) -> Callable:
         """
         def wrapper(*args, **kwargs):
             chat_id = kwargs.pop("chat_id", None)
-            message = {"role": "user", "content": kwargs.pop("message")}
+            # system messages stuff
+            system_kwarg = kwargs.pop("system", None)
+            if system_kwarg is not None:
+                system_message = {"role": "system", "content": system_kwarg}
+            else:
+                system_message = {"role": "system", "content": config.get("DEFAULT_SYSTEM_MESSAGE")}
```
TheR1D, line 93:
```suggestion
            role = kwargs.pop("system", config.get("DEFAULT_SYSTEM_MESSAGE"))
            system_message = {"role": "system", "content": role}
```

sgpt/make_prompt.py
```diff
@@ -79,8 +82,14 @@ def os_name() -> str:
     if not question.endswith("?"):
         question += "?"
     # TODO: Can be optimised.
-    return SHELL_PROMPT.replace("{shell}", shell).replace("{os}", os) + question
+    return SHELL_PROMPT.replace("{shell}", shell).replace("{os}", os), question
 
 
-def code(question: str) -> str:
-    return CODE_PROMPT + question
+def code(question: str) -> {str, str}:
```
TheR1D, line 88:
```suggestion
def code(question: str) -> Tuple[str, str]:
```

sgpt/app.py
```diff
@@ -52,6 +54,7 @@ def main(
     prompt: str = typer.Argument(None, show_default=False, help="The prompt to generate completions for."),
     temperature: float = typer.Option(1.0, min=0.0, max=1.0, help="Randomness of generated output."),
     top_probability: float = typer.Option(1.0, min=0.1, max=1.0, help="Limits highest probable tokens (words)."),
+    system: str = typer.Option(None, help="Prompt portion for system role."),
```
TheR1D, line 57: This name is a bit confusing, rename it as `role` please.

TheR1D (changes requested):
Hi, Good job, solid pull request! I left couple comments.

sgpt/cache.py
```diff
@@ -84,16 +85,24 @@ def __call__(self, func: Callable) -> Callable:
...
                 return func(*args, **kwargs)
-            kwargs["message"] = self._read(chat_id)
-            kwargs["message"].append(message)
-            response_text = func(*args, **kwargs)
-            kwargs["message"].append({"role": "assistant", "content": response_text})
-            self._write(kwargs["message"], chat_id)
-            return response_text
+            else:
+                kwargs["messages"] = self._read(chat_id)
+                kwargs["messages"].append(system_message)
```
TheR1D, line 100: Check if chat session file not eixsts with given `chat_id` and add system_role just once for initianting a chat session. This is important since after couple of messages we will be passing redundant payloads with useles repeating system role messages. It adds latency to request + response time and OpenAI will spend time figuring out all these role messages. If user passed `--role` with `--chat {id}`:
```python
raise typer.BadParameter("Cannot use --role with initiated chat session.")
```

sgpt/app.py
```diff
@@ -77,15 +80,19 @@ def main(
         prompt = get_edited_prompt()
 
     if shell:
+        if system is not None:
```
TheR1D, line 83:
```suggestion
        if not system:
```

sgpt/app.py
```diff
@@ -77,15 +80,19 @@ def main(
...
+            raise typer.BadParameter("Cannot use --system with --shell.")
         # If default values, make response more accurate.
         if top_probability == 1 == temperature:
             temperature = 0.4
-        prompt = make_prompt.shell(prompt)
+        system, prompt = make_prompt.shell(prompt)
     elif code:
-        prompt = make_prompt.code(prompt)
+        if system is not None:
```
TheR1D, line 90:
```suggestion
        if not system:
```

sgpt/cache.py
```diff
@@ -84,16 +85,24 @@ def __call__(self, func: Callable) -> Callable:
...
-                kwargs["message"] = [message]
+                kwargs["messages"] = [system_message, message]
                 return func(*args, **kwargs)
-            kwargs["message"] = self._read(chat_id)
-            kwargs["message"].append(message)
-            response_text = func(*args, **kwargs)
-            kwargs["message"].append({"role": "assistant", "content": response_text})
-            self._write(kwargs["message"], chat_id)
-            return response_text
+            else:
```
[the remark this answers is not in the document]
lpurdy01: I know, I just like the syntactic sugar, it feels more readable to me for every `if` to have an `else` even if there are returns in between. :) I'll just switch it to a commented out else.

README.md
```diff
@@ -232,7 +242,7 @@ REQUEST_TIMEOUT=60
 │ --cache                                       Cache completion results. [default: cache]                  │
 │ --animation                                   Typewriter animation. [default: animation]                  │
 │ --spinner                                     Show loading spinner during API request. [default: spinner] │
-│ --help                                        Show this message and exit.                                 │
+│ --help                                        Show this messages and exit.                                 │
```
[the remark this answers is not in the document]
lpurdy01: Yep, it was just a overeager auto refactor

README.md
```diff
@@ -125,6 +125,15 @@ python fizz_buzz.py
 # ...
 ```
 
+### Custom System Roles
+You can use the `--system` option to specify a custom system role. This is useful to tell the model how it should act
+along with what rules you want it to follow.
+
+```shell
+sgpt --system "You are a json file generator. Only ever return valid json syntax." "What are your rules?"
```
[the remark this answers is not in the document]
lpurdy01: I was hoping it would respond with its rules in .json, but gpt-3.5 doesn't appear to be smart enough to get it.

sgpt/cache.py
lpurdy01: I like the idea of `raise`ing in the event of a user trying to add another system role message. That is a good solution. Because --shell and --code are also moving to system role messages the error prompt can't be that simple though.

sgpt/cache.py
```diff
@@ -83,22 +86,32 @@ def __call__(self, func: Callable) -> Callable:
         :param func: The chat function to cache.
         :return: Wrapped function with chat caching.
         """
+
         def wrapper(*args, **kwargs):
-            chat_id = kwargs.pop("chat_id", None)
-            role = kwargs.pop("system", config.get("DEFAULT_SYSTEM_MESSAGE"))
+            # role stuff
+            role = kwargs.pop("role", None)
+            if not role:
```
lpurdy01, line 93: This `if not role:` syntax is necessary, because the auto *replacement if not exists* of `kwargs.pop` doesn't work because `'role'` will always be passed. (Even it what is passed is just a `none`.) 

It works the same way as chat_id.

sgpt/app.py
lpurdy01: `if not role` triggers by default because `role` is by default `None`. I went with `if role:`

lpurdy01 (commented):
I think this addresses all previous suggestions.

sgpt/make_prompt.py
```diff
@@ -79,8 +82,14 @@ def os_name() -> str:
     if not question.endswith("?"):
         question += "?"
```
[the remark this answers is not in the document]
lpurdy01: Please check if this was resolved how you were expecting :)

sgpt/cache.py
```diff
@@ -82,18 +86,32 @@ def __call__(self, func: Callable) -> Callable:
         :param func: The chat function to cache.
         :return: Wrapped function with chat caching.
         """
+
         def wrapper(*args, **kwargs):
+            # role stuff
+            role = kwargs.pop("role", None)
```
TheR1D, line 92:
```suggestion
            role = kwargs.get("role") or config.get("DEFAULT_SYSTEM_ROLE")
```

sgpt/cache.py
```diff
@@ -82,18 +86,32 @@ def __call__(self, func: Callable) -> Callable:
...
+            if not role:
+                role = config.get("DEFAULT_SYSTEM_ROLE")
```
TheR1D, line 94:
```suggestion
            role = kwargs.pop("role", None) or config.get("DEFAULT_SYSTEM_ROLE")
```

lpurdy01: This appears to work and is more concise! 

Want to apply this to how chat_id checks if it exists as well?

lpurdy01:
I want to add `--save-role <role_name>`, `--recall-role <role_name>`, `--list-roles`, and `--show-role` operators, but I think we should have an issue chain to see if that is how people want it implemented and might be best in a different pull request.

## To be used like

How to save a role as a custom role:
```
$ sgpt --role "You are a terminal emulator. Respond as if prompts received are directly executed on a bash shell" --save-role virtual_shell 

Custom role 'virtual_shell' saved:
"You are a terminal emulator. Respond as if prompts received are directly executed on a bash shell"
```
How to use a custom role:
```
$ sgpt --recall-role virtual_shell "ls -la"

file1.txt
Documents
...
```
How to view available roles:
```
$ sgpt --list-roles

Base roles:
'default'
'shell'
'code'

Custom roles:
'virtual_shell'
```
How to see what a role is:
```
$ sgpt --show-role virtual_shell

Custom role 'virtual_shell':
"You are a terminal emulator. Respond as if prompts received are directly executed on a bash shell"
```
How to update a role:
```
$ sgpt --role "You are a terminal emulator. Respond as if prompts received are directly executed on a PowerShell" --save-role virtual_shell 

Custom role 'virtual_shell' saved:
"You are a terminal emulator. Respond as if prompts received are directly executed on a PowerShell"
```

TheR1D:
I was testing these changes today, and seems current GPT-3.5 model doesn't pay much attention to `system` role, as it is described in OpenAI [documentation](https://platform.openai.com/docs/guides/chat/introduction):
>gpt-3.5-turbo-0301 does not always pay strong attention to system messages. Future models will be trained to pay stronger attention to system messages.

I was asking for generating few shell commands, and it doesn't perform well while generating clean shell commands. For example `what is in current folder` will return a command along with description which we don't want.

Last couple days I spent a lot of time to find best performing prompt to get a clean shell command, even for complicated queries. With few tricks and without defining huge list of rules I managed to make a prompt which most of the time getting clean shell comamnd or code (currently on main). Could you test these new changes with system role? If it will perform well we can merge this feature.

lpurdy01:
Don't review yet. Just testing the force pushing. 
I just re-engineered how roles work in support of #110 and my desires to have custom roles be savable. Currently working on testing, debugging, documentation, and refinement.

TheR1D:
I am planning to merge my PR https://github.com/TheR1D/shell_gpt/pull/130 soon. You will need to rebase it once more. However, this should not be difficult since I notice that you are not modifying core modules.

TheR1D:
Could you please rebase your branch with `main`. I will try to find a solution for your issue with `ChatSession` (aka `ChatCache`) later today.

lpurdy01:
Working on it.

lpurdy01:
It is very annoying to get black and pylint to agree.

lpurdy01:
It is currently failing `integrational_tests.test_code` because it is adding markdown syntax to the code response, and `integrational_tests.test_validation_code_shell` because this test isn't relevant as with the `--role` system you can't put two arguments in at the same time.

sgpt/app.py
```diff
@@ -78,37 +92,47 @@ def main(  # pylint: disable=too-many-arguments
...
+        prompt = get_edited_prompt(prompt)
 
     api_host = config.get("OPENAI_API_HOST")
     api_key = config.get("OPENAI_API_KEY")
     client = OpenAIClient(api_host, api_key)
 
     if chat:
-        full_completion = ChatHandler(client, chat, shell, code).handle(
-            prompt,
+        # TODO: Why doesn't execution continue after .handle returns?
```
lpurdy01, line 111: On my system execution doesn't continue after `Handler.handle()` returns. It looks like it goes into some kind of debug trace, but no error message is output to the terminal. 

In any case this stops the prompting for "Execute shell command" on chats.

sgpt/role_manager.py
```diff
@@ -0,0 +1,221 @@
...
+        "test",
+        "Hello, You are a role testing AI."
+        "Please always respond with a more enthusiastic test response, "
+        "and all context provided.",
+        executable_returns=False,
+        prompt_structure="Hello, {prompt}!",
+        conversation_lead_in=[
+            {"role": "user", "content": "{shell} and os {os} Test Test"},
+            {"role": "assistant", "content": "{shell} and os {os} Test Test Test!"},
+        ],
```
lpurdy01, line 215: I get a typing error here of `Expected type 'Optional[list[Mapping[dict, dict]]]', got 'list[Union[dict[str, str], dict[str, str]]]' instead` 

Not sure how to correct this, because it should accept any list of dicts with the correct content.

sgpt/handlers/chat_handler.py
```diff
@@ -126,52 +122,45 @@ def show_messages(cls, chat_id: str) -> None:
...
+                # Error message about how the user is asking for a different
+                # system chat than the one that already exists
                 raise BadArgumentUsage(
-                    f'Chat "{self.chat_id}" was initiated as code assistant, '
-                    "and can be used with --code only"
+                    f"Chat id:'{self.chat_id}' was initiated with system role as"
+                    f" \n'{existing_system_prompt}'\n\n "
+                    f"Can't be continued with new system"
+                    f" prompt \n'{hypothetical_messages[0]['content']}'"
                 )
```
lpurdy01, line 150: This is currently how I am handling the existing chat history being for a different role. It would be really nice to have some variable saved into the chat history so the user wasn't confronted with a wall of text whenever they enter the wrong role for a chat. 

This would also make it much easier to identify what chat was used, so potentially Shell GPT could automatically select the correct role. (Bearing in mind that role files may change.)

TheR1D:
Sry, I did a force push since I was chaging a lot of logic regarding to roles. We can't affect any core features of `sgpt` like `--code` and `--shell`, they should stay, we can't just remove them. I run more tests and seems current GPT-3.5 model doesn't really care about system role. In more cases it is just acting like there was not any system messages (especially with low temperatures like `0.1`). I think we should wait with this feature until GPT-4 will be available for every one. I have pushed one commit where outlined some crucial logic, and made it easy to implement.

lpurdy01:
Np, I just have to figure out how to pull that into my version.

I know GPT-3.5 doesn't pay attention to the system role. That was why I implemented the `PROMPT_STRUCTURE` key. For shell and code it delivers completions to the API the same as what is currently in main (with just an added system prompt for gpt-3.5 to ignore). 

I don't think the role system should be just the `system` role, because when I was playing around (even with gpt-4) on #108 the system role wasn't enough to explain to the model what it was and what it could do. It took some extra prompt structuring and conversation lead in before it understood that it could ask to execute commands and see the result. 

I also think the role system should be more complex to avoid community separation from this project getting forked for specialized purposes #110 . It would be better if Shell GPT was an extensible foundation with a user interface where people wanted it (in their terminal). To support this, the role system will need the complexity where shell and code can be easily included as other types of roles. 

To avoid repeated edge case detection in `ChatHandler` and `DefaultHandler` shell and code should be removed as arguments. Then they can be specified as just another type of role, but if you want the `--shell` `-s` and `--code` arguments to still exist they could be handled in app.py with something like:
```
    # Check if --shell or --code is called and set role accordingly while only allowing 
    # one of the three to be called
    # If role and shell or role and code are called, error
    if role and (shell or code):
        raise typer.BadParameter("Cannot use --role and --shell or --code together.")
    # If neither is called, then the role is set to the default role
    if shell and code:
        raise typer.BadParameter("Cannot use --shell and --code together.")
    
    if shell:
        role = "shell"
    
    if code:
        role = "code"
```
I think it would be weird to have different layers of `roles` where some like `shell` or `code` are defined in python while other custom ones are in files. I think it would be better to make a fully extensible system. 

Finally, the reason I want to implement this role system is because I want to get back to building a semi autonomous system administration and coding assistant, and Shell GPT could become that. I want to have a way to 'pre-train' a model in conversation about things like: 
[Introducing Autonomous Task Orchestration Manager (ATOM) Framework](https://github.com/daveshap/ATOM_Framework)
[Heuristic Imperatives](https://github.com/daveshap/HeuristicImperatives)
[Rolling Episodic Memory Organizer (REMO) for autonomous AI systems](https://github.com/daveshap/REMO_Framework)

And also teach it what interfaces it has access to with the system to understand its environment. 

To do this the role system needs to be more complicated than just exposing the `system` role. It needs to be a customizable conversation initiation system and prompt structuring database. At that point why not just role `shell` and `code` inside?

lalligagger:
I figured I'd chime in as I've thought some more about my use case/ similar interest in an extensible ShellGPT.

1. I think that the default `--code` and `--shell` roles are the right ones to introduce users to sgpt functionality, and especially the `-se` interface is something I'd like to keep using and appreciate it's robustness for CLI command generation.
2. My test cases so far have been adding more to the "context window" ahead of what the user types as their prompt. E.g., after the default shell instructions there is some info about CLI tools 1,2,..,N and their syntax. This has enabled me to use natural language to generate mostly accurate commands in the tools of choice, which are too niche for GPT3.5 and even GPT4 to reliably get correct. I agree that we can't just use the system role, at least for GPT3.5, the user role context needs to be added as well. (Maybe this is the main drift from the original intent of this PR?)
3. As far as I've gotten with the project, I feel like I will probably want to do similar with the `--code` role. I.e., the desired output of commented python code with no other returned context/ commentary is the same. However it would also reference specific documentation that I've provided in the context window.
4. I also think it's interesting to modify the DEFAULT_SYSTEM_MESSAGE above. So, the CLI commands and flags to sgpt would be exactly the same, but _somewhere_ I can configure custom context/ roles for each of: "chat," "code," "shell" so that my desired info is included.

So most of what @lpurdy01 describes here would be immediately useful to me, but I also agree on keeping the current CLI structure as it is both a sensible jumping off point and at least in my case the desired long-term usage.

In fact, I could probably get by with just these 3 custom role types if context windows were unlimited. There is a chance I'd create multiple roles for sub-domains (one is an image retrieval expert, one plans satellite trajectories, etc.) however it looks like the slick way to do it would be embeddings & a layer that adds only the needed context to the window during prompt generation. If this first step has basic ability for saving/ retrieving custom roles and pointing them to one of the three sgpt archetypes I would be very happy! :)

lpurdy01:
@lalligagger Thank you much for jumping in! 
I like the auto context window idea. I have been wanting to do something like that including system context (like the result of ls or cat commands), but I haven't hypothesized a way to implement it yet. 

In the last force push some context got wiped out. 
Have a look at: https://github.com/TheR1D/shell_gpt/blob/892380185f33ea2178e54d4c962153af95301ead/README.md#the-role-system

And: https://github.com/TheR1D/shell_gpt/blob/892380185f33ea2178e54d4c962153af95301ead/README.md#custom-system-roles

lpurdy01:
This is what my shell.json looks like:
```
{
    "EXECUTABLE_RETURNS": true,
    "SYSTEM_MESSAGE": "\nYou are Command Line App ShellGPT, a programming and system administration assistant.\nYou are managing {os} operating system with {shell} shell.\nProvide only {shell} commands for {os} without any description.\nIf there is a lack of details, provide most logical solution.\nEnsure the output is a valid shell command.\nIf multiple steps required try to combine them together.\n",
    "PROMPT_STRUCTURE": "###\nProvide only {shell} commands for {os} without any description.\nIf there is a lack of details, provide most logical solution.\nEnsure the output is a valid shell command.\nIf multiple steps required try to combine them together.\nPrompt: {prompt}\n###\nCommand:"
}
```

lpurdy01:
@TheR1D I can work on rebasing this capability using the outline you provided today, but I don't want to have both of us working on a duplicate implementation. Where are you at with this?

TheR1D:
> @TheR1D I can work on rebasing this capability using the outline you provided today, but I don't want to have both of us working on a duplicate implementation. Where are you at with this?

Well I already have working code https://github.com/TheR1D/shell_gpt/pull/183, but big part of this feature is acutally to test it properly.

TheR1D:
https://github.com/TheR1D/shell_gpt/pull/183

Issue #54, opened 2023-03-06
Feature request: Conversation loop mode


buzzert:
I think it would be really cool if there was an option to start `sgpt` with a loop/readline mode that continuously accepted input and automatically generated a chat identifier. 

For example, you should be able to invoke something like `sgpt --chat-continuous`, which would then print some kind of prompt character (e.g., ">"), and the user can interact with ChatGPT just by typing and hitting enter, rather than having to format the entire input as a command line argument.

TheR1D:
It can be implemented in coming patches. Using CLI options for `--chat` is not the best, all chat functionality should be moved to dedicated subcommand, something like:
```
sgpt chat [id]  # Loop mode
sgpt chat [id] --no-chat  # Normal mode.
sgpt chat [id] --show  # Show all message related on session.
sgpt chat --list  # Show sessions.
```

PR #94, opened 2023-03-25
Implementation of interactive mode

TheR1D:
This pull request adds a new interactive mode to `sgpt`, which can be accessed using the `--interactive` or `-i` command line options. With this feature, users can interact with sgpt in a more dynamic way, and it works seamlessly with both the `--shell` and `--code` options. This makes it possible to refine shell commands or code iteratively, leading to a more efficient workflow. This update addresses issues https://github.com/TheR1D/shell_gpt/issues/54 and partialy https://github.com/TheR1D/shell_gpt/issues/78. Attaching gif demo to the pull request description.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import setup, find_packages
 # pylint: disable=consider-using-with
 setup(
     name="shell_gpt",
-    version="0.8.2",
+    version="0.8.3",
     packages=find_packages(),
     install_requires=[
         "typer~=0.7.0",
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ A command-line productivity tool powered by OpenAI's ChatGPT (GPT-3.5). As devel
 
 ## Installation
 ```shell
-pip install shell-gpt==0.8.2
+pip install shell-gpt==0.8.3
 ```
 You'll need an OpenAI API key, you can generate one [here](https://beta.openai.com/account/api-keys).
 
@@ -121,7 +121,7 @@ python fizz_buzz.py
 ```
 
 ### Chat
-To start a chat session, use the `--chat` option followed by a unique session name and a prompt:
+To start a chat session, use the `--chat` option followed by a unique session name and a prompt. You can also use "temp" as a session name to start a temporary chat session.
 ```shell
 sgpt --chat number "please remember my favorite number: 4"
 # -> I will remember that your favorite number is 4.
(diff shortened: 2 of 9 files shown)


setup.py
```diff
@@ -3,7 +3,7 @@
 # pylint: disable=consider-using-with
 setup(
     name="shell_gpt",
-    version="0.7.3",
+    version="0.7.4",
```
joeytwiddle, line 6: Just a thought. If you wanted to avoid typing `r<Enter>` each time, you could instead go into revision mode immediately, but look out for special commands:

```
:e --- execute
:q --- quit
:* --- error message (: is always considered a command)
```

Disadvantage: You would not be able to send a reply to GPT starting with `:`

Maxattax97:
I badly want this feature, but not just for code. It'd be very nice to just have back and forth chat (no revision, execute, etc) in what I would call a "[REPL](https://en.wikipedia.org/wiki/Read%E2%80%93eval%E2%80%93print_loop)".

Anything I can do to help?

TheR1D:
@Maxattax97 Would be nice if you could test it before it will go to main :)

TheR1D:
Few cool improvements:
* Now `--chat` will show chat history right away (`--show-chat` will be removed)
* If used doesn't pass prompt, it will ask for input.
* Interactive loop mode allows to execute shell commands or just refine prompt (works with normal chat, shell, code).
* Few "UI" elements to make it more readable.

Maxattax97:
Okay I've tried it on my system, it works great.
- making interactive by default may break scripts that people have ([I have one such script](https://github.com/Maxattax97/miscellaneous/blob/master/config/rofi/menu/chatgpt#L9))
- there are no flags for disabling interactive mode in `--help`
- nitpick: `Enter your prompt:` might be okay, but getting `Prompt:` after each line feels kind of verbose. Could we just lead with `$`, `>`, `>>>` (Python style), `]`, `)`, ... lots of options that offer brevity that make it like a REPL which makes terminal people feel at home
- nitpick: the pink is kind of bright and annoying to read for lots of text... green? blue? Or maybe we lead with `<` or some kind of character to denote that ChatGPT said it (although extra character may be a concern for piping I/O...)

Maxattax97:
Also, what happened to `--no-animation`? I loved that option...

TheR1D:
@Maxattax97
It will not break your script, interactive mode is default for `--chat` mode only. Animation and spinner options were removed since now we are streaming results from OpenAI right to stdout, this allows to get response word by word instanly after request was initiated.

SoarinFerret:
I have started testing / using this on my linux desktop in a dedicated terminal - works great!

Some quality of life things would be nice to add (I have already added these in my personal fork):
* Adding `import readline` to the top of `app.py` - this allows users to cycle through previous prompts and edit them.
* Adding an "exit" keyword to close the app. Even though it says use ctrl-c, I have found trying to use `exit` multiple times

TheR1D:
@SoarinFerret Thank you for the feedback. In the latest commit, there is an option to quite REPL mode using `exit()`. I will add suggested `readline` import, make sense.

PR #100, opened 2023-03-26
install requires more flexible

wux345:


.gitignore
```diff
@@ -0,0 +1,126 @@
...
+
+# PyCharm
+*.pyc
+.idea/
+.idea/workspace.xml
+.idea/tasks.xml
+.idea/dictionaries/
+.idea/codeStyles/
+.library/
+__pycache__
```
joeytwiddle, line 126: Great. I also need this, and the `*.egg-info/` entry

setup.py
```diff
@@ -1,16 +1,22 @@
...
 
+
+def get_requires():
+    requires = []
+    with open("requirements.txt") as requirements_file:
+        libs = requirements_file.readlines()
+        for lib in libs:
+            if not lib.startswith("setuptools"):
+                requires.append(lib.strip())
+    return requires
```
joeytwiddle, line 11: Clever. But perhaps there is an easier way. Do you trust this guy?

```py
$ sgpt 'In Python, what is the recommended way to read and parse requirements.txt, for use in setup.py?'

The recommended way to read and parse requirements.txt in Python for use in setup.py is to use the
`parse_requirements` function from the `pip` package. 

Here's an example of how to use it:

from pip.req import parse_requirements
from setuptools import setup, find_packages

# Parse the requirements file
requirements = parse_requirements('requirements.txt', session='')

# Get the list of requirements as strings
install_requires = [str(r.req) for r in requirements]

# Setup.py configuration
setup(
    name='my_package',
    version='1.0',
    description='My package description',
    packages=find_packages(),
    install_requires=install_requires,
)
```

joeytwiddle: Curiously, if I add `--code` then the robot suggests `f.read().splitlines()` instead of `parse_requirements()`

wux345: Hi, nice to get your feedback, When using parse_requirements function, it is important to consider the compatibility with different versions of Python and pip. In some cases, the function may not work as expected if the versions of Python and pip are not compatible.
```python
try: # for pip >= 10
    from pip._internal.req import parse_requirements
except ImportError: # for pip <= 9.0.3
    from pip.req import parse_requirements
```
my local pip 23.0.1, it's work below code
```python
def get_requires():
    requires = [re.requirement for re in parse_requirements('requirements.txt', '')]
    return requires
```
if i use `re.req` where you show by sgpt, i will get a problem which has no attribute req.

joeytwiddle: Thanks, that's interesting to know.

.gitignore
TheR1D: This very specific for each user. Usually using "personal" `.gitignore` is enought.
```shell
git config --global core.excludesFile ~/.gitignore
```

setup.py
TheR1D: `requirements.txt` might include packages for testing or development specific packages, which we don't need in distribution package.

TheR1D:
Thank you for the PR.
I left couple comments above. And we will be migrating to [TOML](https://toml.io/en/) configs soon. I don't think parsing and including packages from requirements to setup is a best idea.

Issue #101, opened 2023-03-26
Use an example in the prompt, to encourage ChatGPT to behave


joeytwiddle:
Sometimes ChatGPT is disobedient. For example: #99 But I don't know how common this is. Do you see this often? (I think I might have seen it once.)

Due to GPT's predictive nature, it was recommended to prompt it with an incomplete conversation, so that it would complete the rest. Although ChatGPT is different, it may respond to the same treatment.

This approach (shown below) is taken by [`plz`](https://github.com/zahidkhawaja/rusty/blob/c74a7439c40153e5f0f60120ae7f5d5d4d18fbe9/src/main.rs#L42-L63), while [`rusty`](https://github.com/m1guelpf/plz-cli/blob/9d4549ba972a3b0d03533dced6539c6f45b3d5fd/src/main.rs#L147) passes `#!/bin/bash\n` to initialise the response.

### Suggestion

When we are asking ChatGPT to generate some shell code, we may want to give it an example first.

```
{The current shell prompt rules}

Input: List the files in the root folder
Output: ls /

Input: ${user_question_here}
Output: 
```

For the code prompt, even though we don't know the user's target language, it might still be possible to do something similar.

joeytwiddle:
I have put a solid proposal in #102, to better share the idea. Feel free to close it.

TheR1D:
Prompts improvements were inmplemented in https://github.com/TheR1D/shell_gpt/pull/107 now `sgpt` performs better when making suggestions for `--code` or `--shell`

PR #102, opened 2023-03-26
Prompt with example

joeytwiddle:
This is an example of the suggestion in issue #101, made into a PR to clarify.

This probably needs a lot of testing before we can be sure its beneficial.

(If we want to be scientific, we would run [a bunch of queries](https://github.com/shorwood/gptsh/blob/7dab245edf1d37282a9c49938fd74eff3dd9e488/data/promptTemplate.txt) over the old and new prompts, perhaps multiple times, and compare results.)

Concerns:

1. Although the examples are not multi-line (input or output) the real question or answer could be multi-line

Feel free to close. Just sharing ideas this way.

### DIFF
--- a/sgpt/make_prompt.py
+++ b/sgpt/make_prompt.py
@@ -43,7 +43,15 @@ This is important you MUST follow the above rules.
 There are no exceptions to these rules.
 You must always follow them. No exceptions.
 
-Request: """
+For example:
+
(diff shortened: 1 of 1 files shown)


TheR1D:
Thank for the PR, prompts were redesigned in the last patch, now they are working much better when providing shell or code completion.

PR #104, opened 2023-03-26
added --reset-key and black style badge

Soycid:
fixed issue #30

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,10 @@
 # Shell GPT
+
+[![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
+
+
 A command-line productivity tool powered by OpenAI's ChatGPT (GPT-3.5). As developers, we can leverage ChatGPT capabilities to generate shell commands, code snippets, comments, and documentation, among other things. Forget about cheat sheets and notes, with this tool you can get accurate answers right in your terminal, and you'll probably find yourself reducing your daily Google searches, saving you valuable time and effort.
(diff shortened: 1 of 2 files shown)


TheR1D:
Thank you for the PR.
These changes do not address all the suggestions mentioned in the issue.

Soycid:
> Additionally impement API key validation. Since OpenAI doesn't have dedicated endpoint just for key validation, call https://api.openai.com/v1/completions with some test prompt, make sure response status code is not 403, if so, ask user for valid API key.

ill do this later. ty for pointing it out.

maxsu:
Here is a sketch for a key validator:

```py
import openai

def is_api_key_valid(key):
    try:
        openai.Completion.create(
            api_key=key,
            engine="davinci",
            prompt="This is a test.",
            max_tokens=1
        )
    except:
        return False
    else:
        return True

# Check the validity of an API key
api_key = 'YOUR_API_KEY'
api_key_valid = is_api_key_valid(api_key)
print("API key is valid:", api_key_valid)
```

This should work because the completion api create method [calls its super method](https://github.com/openai/openai-python/blob/1be14ee34a0f8e42d3f9aa5451aa4cb161f1781f/openai/api_resources/completion.py#L25).  This method  [allows us to  pass an api key](https://github.com/openai/openai-python/blob/1be14ee34a0f8e42d3f9aa5451aa4cb161f1781f/openai/api_resources/abstract/engine_api_resource.py#L128).

Note I haven't tested or integrated this yet.

PR #124, opened 2023-03-31
Updated Dockerfile

karlisk:
Overall - fewer steps, cleaner looking Dockerfile

- Using `/home/app` as app directory and installing shell_gpt in user install mode oppose to installing system-wide in the container
- Adding initial shell_gpt config as ENV variables
- Setting multiple ENV params in a single step oppose to individual steps during Docker build
- Using full path to executables avoids potential errors like: `Error: crun: executable file `sgpt` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found`

Unsure why is this needed but I'm assuming it's for debugging/development purposes
```
# Copy all files into the container
COPY . /home/app
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ USER app
 WORKDIR /home/app
 
 # Copy all files into the container
-COPY . /tmp/test
+COPY . /home/app
 
 # Set up Environment parameters
 # These are overriden if any of the variables are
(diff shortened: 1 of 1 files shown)


TheR1D:
Thank you for the PR.
We can't build docker image using shell-gpt from PYPI registry, it should be builded from source. The `Dockerfile` is also used in [pipeline](https://github.com/TheR1D/shell_gpt/blob/main/.github/workflows/docker.yml), in this case it will be building images using latest version from pypi while it could be not published yet.

PR #130, opened 2023-04-01
Refactor, more tests, lint fixes, optimise

TheR1D:
Initially, I didn't design ShellGPT to handle the numerous features it now supports. Over the past two weeks, we've implemented several new features, which has resulted in a somewhat cluttered codebase. This PR addresses this issue by introducing additional integration tests, fixing linting (which was broken), refactoring Python modules, and creating a more organized project structure. These changes should make it easier to implement new features in the future.

### DIFF
--- a/.github/workflows/lint_test.yml
+++ b/.github/workflows/lint_test.yml
@@ -1,4 +1,4 @@
-name: Run linters and unittests
+name: Lint and Test
 
 on:
   push:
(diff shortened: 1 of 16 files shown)


lpurdy01:
I haven't wanted to touch the chat cache system in my pull request #92, because the change I need would probably break all previously cashed chats. 

The system of determining if something was a code or shell chat by:
```
self.is_shell_chat = chat_history and chat_history[0].endswith("###\nCommand:")
self.is_code_chat = chat_history and chat_history[0].endswith("###\nCode:")
``` 
is messy and not extensible to custom roles. 

If you could add a variable to ChatCache files that stores what role a chat was started with. It would be super helpful. Because right now my solution is to just print out the entire system message requested and the one stored in the chat cache file and force the user to spot the differences without any context. 

I can also add this after your refactor if you think it would be ok, but since you are messing around in there thought I would ask.

TheR1D:
@lpurdy01 Once this changes will be merged in to `main`, please rebase your feature branch. We will then work on resolving the issue in your branch.

PR #149, opened 2023-04-06
Native $SHELL commands and PowerShell fix

TheR1D:
* Native $SHELL command execution instead of default `/bin/sh`.
* Added dependency `pyreadline3` for Windows systems.
* Fixed a bug when `sgpt` couldn't recognize `powershell` or `cmd` on Windows.
* Better integration with `powershell` and `cmd` on Windows.


xihuai18:
Powershell executable is named pwsh.exe instead of powershell.exe in Powershell 7, I think retrieving the parent processes is a better solution, as in #141

PR #151, opened 2023-04-07
Model choice option to support GPT-4

TheR1D:
`sgpt --model gpt-4 "test prompt"`

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -29,7 +29,7 @@ def main(  # pylint: disable=too-many-arguments,too-many-locals
         help="The prompt to generate completions for.",
     ),
     model: ModelOptions = typer.Option(
-        ModelOptions(config.get("DEFAULT_MODEL")),
+        ModelOptions(config.get("DEFAULT_MODEL")).value,
         help="OpenAI GPT model to use.",
     ),
     temperature: float = typer.Option(
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -29,7 +29,7 @@ def main(  # pylint: disable=too-many-arguments,too-many-locals
         help="The prompt to generate completions for.",
     ),
     model: ModelOptions = typer.Option(
-        ModelOptions.GPT3.value,
+        ModelOptions(config.get("DEFAULT_MODEL")),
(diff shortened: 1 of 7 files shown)


sgpt/app.py
```diff
@@ -19,15 +19,19 @@
...
 
-def main(  # pylint: disable=too-many-arguments
+def main(  # pylint: disable=too-many-arguments,too-many-locals
     prompt: str = typer.Argument(
         None,
         show_default=False,
         help="The prompt to generate completions for.",
     ),
+    model: ModelOptions = typer.Option(
+        ModelOptions.GPT3.value,
```
lpurdy01, line 32:
```suggestion
        ModelOptions(config.get("DEFAULT_MODEL")),
```

lpurdy01:
These tests were performed on #153 

Test gpt-4:
```
sgpt --model gpt-4 "test show me some reasoning"
As an AI language model, I can provide you with an example of reasoning. Let's consider a simple problem-solving scenario:

Problem: John wants to bake a cake, but he doesn't have any eggs.

Reasoning process:

1. Identify the problem: John needs eggs to bake a cake.
2. Consider possible solutions:
   a. Go to the store and buy eggs.
   b. Ask a neighbor for eggs.
   c. Find a substitute for eggs in the cake recipe.
3. Evaluate the solutions:
   a. Going to the store: This option requires time and effort, but it guarantees that John will have eggs.
   b. Asking a neighbor: This option is quicker and requires less effort, but it depends on the neighbor having eggs and being willing to share.
   c. Finding a substitute: This option requires research and might not produce the same results as using eggs, but it allows John to bake the cake without leaving his house.
4. Choose the best solution based on the evaluation: In this case, let's assume John chooses option (a) because he wants to ensure the best results for his cake.
5. Implement the solution: John goes to the store, buys eggs, and bakes the cake.

This is an example of reasoning through a problem to find a solution.
```

Test gpt-3.5-turbo:
```
sgpt --model gpt-3.5-turbo "test show me some reasoning"
I'm sorry, I'm not sure what you mean by "test show me some reasoning". Can you please provide more context or clarify your request?
```

I don't have access to gpt-4-32k

PR #152, opened 2023-04-08
Better prompt with pwd

xihuai18:


setup.py
```diff
@@ -10,6 +10,7 @@
         "requests~=2.28.2",
         "rich==13.3.1",
         "distro~=1.8.0",
+        "psutil",
```
TheR1D, line 13: This should be installed only for Windows users.

sgpt/app.py
```diff
@@ -12,17 +12,17 @@
 
 
 import os
-
 # To allow users to use arrow keys in the REPL.
 import readline  # pylint: disable=unused-import
 
+import psutil
```
TheR1D, line 18: Should be imported only for Windows users.

sgpt/app.py
```diff
@@ -128,8 +128,13 @@ def main(  # pylint: disable=too-many-arguments
...
         )
 
+
     if not code and shell and typer.confirm("Execute shell command?"):
-        os.system(full_completion)
+        # Get the parent process name.
+        shell_name = get_shell_name()
+
+        # print(f'executing {shell_name} -c "{full_completion}"')
+        subprocess.run([shell_name, "-c", full_completion], check=False)
```
TheR1D, line 137: Change it for REPL mode as well.

sgpt/make_prompt.py
```diff
@@ -1,11 +1,12 @@
...
 import platform
-from os import getenv
-from os.path import basename, splitext
 from distro import name as distro_name
+from sgpt.utils import get_shell_name
 
 
 SHELL_PROMPT = """###
 Provide only {shell} commands for {os} without any description.
+The current working directory is {pwd}.
```
TheR1D, line 9: Not sure about this change. Could you provide some use cases for it?

.gitignore
TheR1D, line 1: Using "personal" .gitignore solves this.
```
git config --global core.excludesfile $HOME/.gitignore
```

sgpt/utils.py
```diff
@@ -19,6 +20,18 @@ def get_mode(cls, shell, code) -> "CompletionModes":
...
 
+def get_shell_name() -> str:
+    """
+    Get the name of the shell.
+
+    :return: Shell name.
+    """
+    proc = psutil.Process(os.getppid())
+    return next(
+        p.name() for p in proc.parents() if "sh" in p.name() or "cmd" in p.name()
```
TheR1D, line 31: This will not work with `zsh` or `bash` or other non `sh` shells, it will also crash sgpt.

TheR1D:
Thank you for the PR.
Currently I don't think this is in some workable shape, since it will be crashing every user with non `sh` shells. Please have a look on `tests/integration_tests.py`, and add testing cases for the specifc feature you are implementing.

.gitignore
xihuai18: Learned! thanks

sgpt/utils.py
xihuai18: I am not sure but it works in my zsh shell, the os is Ubuntu.
image

TheR1D: Ubuntu:
```
proc = psutil.Process(pid=459, name='bash', started='19:24:23')
AttributeError: 'Process' object has no attribute 'parents'
```

MacOS:
```
proc = psutil.Process(pid=1473, name='zsh', status='running', started='01:54:35')
StopIteration
```

xihuai18: Okay, I think this may results from that I did add the current proc itself into the list, and I fix it now.

PR #153, opened 2023-04-09
Added default model to config file

lpurdy01:
This adds a `DEFAULT_MODEL` option to the config file. In support of the `--model` option.

### DIFF
--- a/sgpt/config.py
+++ b/sgpt/config.py
@@ -5,6 +5,8 @@ from tempfile import gettempdir
 
 from click import UsageError
 
+from sgpt.utils import ModelOptions
+
 CONFIG_FOLDER = os.path.expanduser("~/.config")
 CONFIG_PATH = Path(CONFIG_FOLDER) / "shell_gpt" / ".sgptrc"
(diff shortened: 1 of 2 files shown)


sgpt/config.py
```diff
@@ -43,6 +47,7 @@ def init() -> None:
         config["CACHE_LENGTH"] = os.getenv("CACHE_LENGTH", str(CACHE_LENGTH))
         config["CACHE_PATH"] = os.getenv("CACHE_PATH", str(CACHE_PATH))
         config["REQUEST_TIMEOUT"] = os.getenv("REQUEST_TIMEOUT", str(REQUEST_TIMEOUT))
+        config["DEFAULT_MODEL"] = os.getenv("DEFAULT_MODEL", str(DEFAULT_MODEL))
```
TheR1D, line 50: This will break `sgpt` for everyone that will upgrade to the new version. All current `sgpt` users already have a config file. This will not update it, it will create a new config file only if it is not exists. I will fix it in my branch after merge.

PR #171, opened 2023-04-11
simplify --show-chat --list-chat to -sc -lc

MartletH:
simplify --show-chat --list-chat to -sc -lc


levid0s (approved):

TheR1D:
Hi @MartletH thank you for interest and effort. In my opinion show/list chat commands are not used very often, and might be used without shortcuts.

PR #191, opened 2023-04-16
Add version

### LINKED ISSUE #161: Check version of ShellGPT and warn user

Optional:

melpeln:
Close #161

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -34,7 +34,7 @@ def main(
     temperature: float = typer.Option(
         0.1,
         min=0.0,
-        max=1.0,
+        max=2.0,
         help="Randomness of generated output.",
     ),
     top_probability: float = typer.Option(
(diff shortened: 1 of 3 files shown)


sgpt/app.py
```diff
@@ -28,6 +30,11 @@ def main(
         show_default=False,
         help="The prompt to generate completions for.",
     ),
+    version: Optional[bool] = typer.Option(
```
TheR1D, line 33: It can be implemented in more elegant way using callbacks, like we have it with couple other [functions](https://github.com/TheR1D/shell_gpt/blob/main/sgpt/app.py#L105).

sgpt/app.py
```diff
@@ -89,6 +96,10 @@ def main(
 ) -> None:
     stdin_passed = not sys.stdin.isatty()
 
+    if version:
+        print(f'ShellGPT version: {__version__}')
```
TheR1D, line 100: Should use `typer.echo(...)`
`black --check` will fail.

sgpt/app.py
TheR1D, line 1: Please also prvoide a test, have look on [test_integration.py](https://github.com/TheR1D/shell_gpt/blob/main/tests/test_integration.py). Make sure to run `sh scripts/lint.sh` and `sh scripts/test.sh`, more info in [CONTRIBUTING.md](https://github.com/TheR1D/shell_gpt/blob/main/CONTRIBUITNG.md#pull-request).

TheR1D (changes requested):

melpeln:
Would you be open to creating a version.py to expose version number?

Issue #194, opened 2023-04-16
A Natural Language Interface to Enhance User-Friendly Command Line Experience


chinarjoshi:
### Discussed in https://github.com/TheR1D/shell_gpt/discussions/173

<div type='discussions-op-text'>

<sup>Originally posted by **chinarjoshi** April 11, 2023</sup>
Natural language to bash is a powerful tool that I believe deserves its own built in interface into the shell rather than a separate program. I propose an interface where users can toggle between natural language and English with a key binding to create a more user-friendly shell. Not to bash bash, but simple tasks should not take research on which program + flag combo will yield the right result with external programs like man and google. It is a more intuitive interface to toggle between a description and its command with a keybinding that invokes the sgpt program, rather than typing *sgpt -s "command"* every time. This is useful if you forget a flag midway through a command or want to figure out what a command does. Developers of any experience, especially people new to the terminal, frequently face this issue; thus, there must be an interface integrated directly with the shell to save both key presses and the cognitive load of another program.

### This interface behaves as such
Makes a key binding that transforms the content of the buffer based on what is currently in it, which has 4 cases:
1. Shell command -> natural language description
2. Natural language -> shell command
3. No text -> natural language description of previous command
4. If there is natural language in the buffer and the user presses enter, then the command is translated to shell before executing. If it is a non-destructive command, it is directly executed. Otherwise, the user is prompted before execution, just like the default behavior.

### This interface comprises of two changes
1. Shell script that is sourced by the users shell rc file that creates a keybinding that toggles the buffer using the sgpt program
2. New "--describe-shell" command that produces a single imperative sentence that describes a given shell command. This addresses  #118 .

Here is a demonstration of the described behavior. I will make a pull request containing these changes.

https://user-images.githubusercontent.com/68311366/232335365-0983ea04-4eb6-4c79-9eef-add798be93da.mp4

chinarjoshi:
This also works well with the caching feature because English commands feel like aliases when they instantly get translated.

PR #195, opened 2023-04-16
Shell integrations

chinarjoshi:
Implementation of #194 . The key-binding is Control+Space, which doesn't conflict on zsh nor bash.

I decided most commands *sgpt* spits out are good enough to execute directly without supervision. This makes the natural language interface really make it feel like an English shell. However, there are two hacks when executing a natural language command directly that should be discussed:
1. The way command_not_found_handler decides if its English is whether the first word is upper cased and there are at least 6 words in the sentence. I figured this because natural language is more verbose than bash, so an arbitrary length threshold is a good heuristic.
2. The way _sgpt_is_destructive_command checks if the command is destructive is manually check if its in the list (rm mv dd chmod sed git mkfs mkswap). There are of course many more destructive commands, and to be safe the script should be checking a wider range of them.

I also added a .gitignore file with GitHub's Python template because __pycache__ files kept being recognized by git.

If you agree with this interface, the next thing to do is update pyproject.toml to prompt the user to update their shell rc to source this shell script, and update documentation for the --describe-shell flag.


TheR1D:
Thank you for the PR, this is awesome! Great idea and implementation! I have few suggestions that could improve UX:

Consider changing PS1 or implement some other indication of modes. When I was using it, I cound't understand in what mode im currently in. I think it can be fixed by implementing something like `ps1_toggle`:
```shell
...

# This is not pretty, just an example. We can modify current PS1.
# Something like current_ps1 + ">>>" ...
_toggle_ps1() {
  if [[ $PS1 == ">>>" ]]; then
    PS1="<<<"
  elif [[ $PS1 == "<<<" ]]; then
    PS1=">>>"
  fi
  zle reset-prompt
}

# Determine the type of shell, and execute the appropriate keybinding commmand
if [[ -n $ZSH_VERSION ]]; then
    zle -N _sgpt_zsh
    zle -N _toggle_ps1
    bindkey $ZSH_KEYBINDING _sgpt_zsh
else
    bind -x "\"$BASH_KEYBINDING\": _sgpt_bash"
fi
```

At some point it acts like completely passthru string, didn't check why it ishappening yet:
```
...
>>> Hello
hello
...
```

sgpt/app.py
```diff
@@ -113,7 +125,7 @@ def main(  # pylint: disable=too-many-arguments,too-many-locals
 
     if repl:
         # Will be in infinite loop here until user exits with Ctrl+C.
-        ReplHandler(client, repl, shell, code).handle(
+        ReplHandler(client, repl, shell, describe_shell, code).handle(
```
TheR1D, line 128: It doesn't really work in REPL mode.
```
>>>sgpt --repl temp --describe-shell
Entering REPL mode, press Ctrl+C to exit.
>>> ls -la
ls -la
>>> echo hello world
echo hello world
```

sgpt/app.py
```diff
@@ -132,7 +144,7 @@ def main(  # pylint: disable=too-many-arguments,too-many-locals
             caching=cache,
         )
     else:
-        full_completion = DefaultHandler(client, shell, code).handle(
+        full_completion = DefaultHandler(client, shell, describe_shell, code).handle(
```
TheR1D, line 147: Doesn't work in chat mode as well.

sgpt/app.py
```diff
@@ -98,6 +105,11 @@ def main(  # pylint: disable=too-many-arguments,too-many-locals
     if shell and code:
```
TheR1D, line 105: I think validation should be performed here, something like:
```python
if sum([code, shell, describe_shell]) == 1:
    ...
```

sgpt/app.py
```diff
@@ -51,6 +51,13 @@ def main(  # pylint: disable=too-many-arguments,too-many-locals
         help="Generate and execute shell commands.",
         rich_help_panel="Assistance Options",
     ),
+    describe_shell: bool = typer.Option(
```
TheR1D, line 54: Could you please add [test_integration](https://github.com/TheR1D/shell_gpt/blob/main/tests/test_integration.py) for this option?

shell_gpt.sh
```diff
@@ -0,0 +1,125 @@
+ZSH_KEYBINDING='^ '
```
TheR1D, line 1: This will not work for Mac users, since it will conflict with system wide hotkey for keyboard language change. MacOS just intercepts it, and it doesn't reach the shell/terminal.

sgpt/app.py
TheR1D: I believe it would be easier to fix it with roles from main, once the branch will be rebased.

shell_gpt.sh
chinarjoshi: I thought that was for Command-Space, not Control-Space. I can't find any mention of Control-Space on [keyboard shortcuts](https://support.apple.com/en-us/HT201236).

TheR1D: Not sure why it is not there. It seems it [depeneds on version](https://www.google.com/search?q=mac+default+langauge+change+hotkey&oq=mac+default+langauge+change+hotkey&aqs=chrome..69i57j33i10i160j33i22i29i30l2.7271j0j7&sourceid=chrome&ie=UTF-8) of macOS

chinarjoshi: Changed the default binding to Alt+Space. Does this still conflict?

sgpt/make_prompt.py
TheR1D, line 1: Thank you, I completely forgot to remove this on last patch.

TheR1D:
I ran few more tests, here is some findings. On mac (zsh) capitalized commands are valid, e.g.:
```shell
Git --version
# git version 2.37.1 (Apple Git-137.1)
```
So if command exists it fails:
```shell
Find all files with .txt extension
# Find: all: No such file or directory
# Find: files: No such file or directory
# Find: with: No such file or directory
# Find: .txt: No such file or directory
# Find: extension: No such file or directory
```
Maybe we could use some special character to detect natural language input. For instance `? what is in current folder`

Another issue, for some reason it triggers `_sgpt_is_destructive_command` for safe commands like:
```shell
Show git version
Execute destructive shell command? [y/N]:
...
```
Also, I would like to keep default shell behaviour for commands that are not exists:
```
foo
# Should throw command not found.
```

shell_gpt.sh
TheR1D: Yes, still doesn't work.

chinarjoshi:
1. I like the idea of English and shell command being directly translatable, and prefixing with some character might disrupt that. Although, that is exactly what [Github Copilot CLI](https://githubnext.com/projects/copilot-cli) is doing, so it may be a good solution. Another solution is making Control+Enter specify that it is English to execute directly.
2. Right now, *destructive command* is pretty dumb and checks if the first word of the command is in a list of destructive commands, i.e. `git push --force` is destructive, thus `git --version` is marked as such. I will write way to enumerate edge cases for commands that are usually destructive but have flags/subcommands that aren't, and vice versa. Do you think this logic should live in a python file for maintainability, or stay in the shell script so everything is in one place?

TheR1D:
In my opinion it should be on shell script side, since `sgpt` asks confirmation when it executes any shell command, also user can see the shell command it self. I think we can even have dedicated branch for the shell script only. On first `sgpt` launch we will prompt user to install shell integration if user has compatible shell (zsh, bash). So we can split this pull request in to 2 parts, first `--describe-shell` will go to the main branch, and second shell script will go to dedicated branch.

chinarjoshi:
Since gpt can produce shell commands with great accuracy, do you think directly executing natural language is a desirable feature so that the shell script can be fully abstracted away?
I think the ideal setup is as such:
* Control+Space: Toggle translate the buffer without executing
* Enter: If the command is not found, then if the command is natural language, translate it and execute. Otherwise return command not found
* Control+Enter: Explicitly translate buffer before executing

Also, I think every single sgpt user uses either bash, zsh, or fish, so there shouldn't be a need for a separate branch. Even if they use sh, the script won't execute anything incompatible if `BASH_VERSION` and `FISH_VERSION` is checked too.

TheR1D:
Control+Space and Control+Enter seems a good solution.
Speaking of branch, my idea was to create a dedicated branch that would make it easier to develop and maintain the sh scripts in it's own branch (there will be no source code of sgpt just sh script). Ofcourse we will add all information on the main readme how to download and isntall it. We can even have a script which will append everything needed to the `.zsh|bashrc`, it could be called using curl `sh -c "$(curl -fsSL https://raw.githubusercontent.com/.../install.sh)"`. I think this feature should be optional.

chinarjoshi:
Can you create a new branch named `shell-integrations` with no files so I can push the shell script to it?

TheR1D:
Sure, here is [shell-integration](https://github.com/TheR1D/shell_gpt/tree/shell-integrations) branch.

chinarjoshi:
This should be ready to merge because it only implements `--describe-shell`. See #242 .

TheR1D (approved):
Looks good to me, thank you.

PR #217, opened 2023-04-26
Enable piping in extensive Data

Zorro909:
I noticed that piping in file data with newlines broke the code responsible for handling pipes.

After a couple of lines (~4-8 if i remember correctly), it stopped and went into interactive mode. The rest of the pipe input was then used line for line, which was obviously not my intent.

To fix this, I changed the code to read all input from the pipe and then manually switch the pipe to stdin (/dev/tty or CON).
This therefore also allows to pipe in data for a REPL session.

Note: I haven't been able to test it on Windows yet, so I'm only 90% sure it will work.

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -111,8 +111,14 @@ def main(
 ) -> None:
     stdin_passed = not sys.stdin.isatty()
 
-    if stdin_passed and not repl:
-        prompt = f"{sys.stdin.read()}\n\n{prompt or ''}"
+    if stdin_passed:
+        input_lines = sys.stdin.readlines()
+        prompt = "".join(input_lines) + "\n\n" + (prompt or "")
+        # Switch to stdin for interactive input
+        if os.name == "posix":
+            sys.stdin = open("/dev/tty", "r")
+        elif os.name == "nt":
+            sys.stdin = open("CON", "r")
 
     if not prompt and not editor and not repl:
         raise MissingParameter(param_hint="PROMPT", param_type="string")


cryolite-ai:
@Zorro909 I think this is a very useful addition - looking at the snippet of code you wrote I had a thought about a prompt security issue which perhaps needs the attention of @TheR1D and possible mitigation which could be integrated here..

Imagine that the piped in material includes text which looks like a prompt.. that external prompt could try to break out of the current execution line by 'overpowering' the appended prompt.. how would the model know which one was the true prompt..? Let's reduce the likelihood of that by giving cues to make clearer the material which is piped by wrapping it..

Can I suggest that there is the option of pre-pipe and a post-pipe padding - it could be something like:

pre-pipe = "PIPED INPUT:" + \n + " -------------" + \n
post-pipe = \n + "---END-OF-PIPE---" + \n + \n

(I'm sure there may be better or more succinct ways of writing it but it needs to be obvious to the model prompted)

now, if people don't mind the risk and want to save some tokens then they should be able to set these to be blank, but perhaps the default should be some safety wrappers.. and perhaps these should be defined in environment variables with default values?

TheR1D:
Hi @maelp, thank you for your interest and effort. Unfortunately, these changes caused issues with `sgpt` in non-interactive environments and affected our [integration tests](https://github.com/TheR1D/shell_gpt/blob/main/tests/test_integration.py). However, I've created a new [PR](https://github.com/TheR1D/shell_gpt/pull/439) that addresses these issues in non-interactive environments. I've also included you as a co-author in the commit to acknowledge your contribution. Thank you again for your effort.

TheR1D:
Hi @Zorro909, thank you for your interest and effort. Unfortunately, these changes caused issues with `sgpt` in non-interactive environments and affected our [integration tests](https://github.com/TheR1D/shell_gpt/blob/main/tests/test_integration.py). However, I've created a new [PR](https://github.com/TheR1D/shell_gpt/pull/439) that addresses these issues in non-interactive environments. I've also included you as a co-author in the commit to acknowledge your contribution. Thank you again for your effort.

PR #221, opened 2023-04-28
Hide API key from Rich tracebacks

### LINKED ISSUE #160: Hide api_key from logs when sgpt crashing
Hide `api_key` value when sgpt generates log after crash. This is a safety issue since some users might share their `api_key` value without checking the actual output.

cosmojg:
`Typer` uses `Rich` to format tracebacks with `show_locals=True` and `locals_hide_dunder=True`. This PR takes advantage of that second parameter to hide the user's API key (stored in `__api_key`) while retaining other local variables. (Issue #160)

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -129,9 +129,7 @@ def main(
     if editor:
         prompt = get_edited_prompt()
 
-    api_host = cfg.get("OPENAI_API_HOST")
-    __api_key = cfg.get("OPENAI_API_KEY")
-    client = OpenAIClient(api_host, __api_key)
+    client = OpenAIClient(cfg.get("OPENAI_API_HOST"), cfg.get("OPENAI_API_KEY"))
 
     role_class = DefaultRoles.get(shell, code) if not role else SystemRole.get(role)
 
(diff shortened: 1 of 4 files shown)


cosmojg:
This PR fixes Issue #160

sgpt/app.py
```diff
@@ -130,8 +130,8 @@ def main(
         prompt = get_edited_prompt()
 
     api_host = cfg.get("OPENAI_API_HOST")
-    api_key = cfg.get("OPENAI_API_KEY")
-    client = OpenAIClient(api_host, api_key)
+    __api_key = cfg.get("OPENAI_API_KEY")
```
TheR1D, line 133: While it might work as "private" property in `OpenAIClient` and `Config` classes, it doesn't works here, since it is a function, I can still see the `__api_key` in tracebacks.

cosmojg: This is fixed in Rich version 13.1.0

cosmojg: Updated the dependency list in `pyproject.toml` to reflect this: [380bd61](https://github.com/TheR1D/shell_gpt/pull/221/commits/380bd612ce59ded923c7d38502f09bbbf9668c09)

TheR1D (approved):
Looks good to me, thanks.

PR #223, opened 2023-05-02
Update maximum and default temperature values to match OpenAI Chat Completion API

eitamal:
### Description

This pull request updates the maximum temperature value from 1 to 2 and the default temperature from 0.1 to 1 to align with the [OpenAI Chat Completion API](https://platform.openai.com/docs/api-reference/chat#chat/create-temperature). However, the change to the default temperature might be considered a **breaking change** for users. I can revert the default temperature change to avoid potential issues if this is a concern.

### Changes

1. Update the maximum temperature value from 1 to 2 in `app.py` and the documentation in `client.py`.
2. Update the default temperature value from 0.1 to 1 in `app.py`.

### Impact

This change ensures the code is consistent with the OpenAI Chat Completion API's temperature settings. However, users who rely on the current default value may experience different results after this update.

### Alternatives

1. Revert the default temperature change and only update the maximum temperature. This will avoid breaking changes for users but won't fully align with the OpenAI Chat Completion API.
2. Introduce a warning message for users about the change in default temperature value, prompting them to set a desired temperature to maintain the previous default explicitly.

--- 

Let me know if you have any feedback or concerns regarding the change, and thanks for creating this fantastic project!

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -34,7 +34,7 @@ def main(
     temperature: float = typer.Option(
         1.0,
         min=0.0,
-        max=1.0,
+        max=2.0,
         help="Randomness of generated output.",
     ),
     top_probability: float = typer.Option(
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -32,7 +32,7 @@ def main(
         help="OpenAI GPT model to use.",
     ),
     temperature: float = typer.Option(
-        1.0,
+        0.1,
         min=0.0,
         max=2.0,
         help="Randomness of generated output.",
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -32,7 +32,7 @@ def main(
         help="OpenAI GPT model to use.",
     ),
     temperature: float = typer.Option(
-        0.1,
+        1.0,
         min=0.0,
         max=1.0,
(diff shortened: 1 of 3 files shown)


sgpt/app.py
```diff
@@ -32,9 +32,9 @@ def main(
         help="OpenAI GPT model to use.",
     ),
     temperature: float = typer.Option(
-        0.1,
+        1.0,
```
TheR1D, line 35: This will break tests, lets keep it as `0.1`.

TheR1D (changes requested):
Also, please change float range in README.md `FLOAT RANGE [0.0<=x<=1.0]`.

eitamal:
I've made the changes you requested 🙂 

To summarise:

### Changes

1. Update the maximum temperature value from 1 to 2 in `app.py` and the documentation in `client.py`.
2. ~Update the default temperature value from 0.1 to 1 in `app.py`.~ Reverted as @TheR1D [suggested](https://github.com/TheR1D/shell_gpt/pull/223#discussion_r1182910803)
3. Update the argument list documentation in `README.md` to reflect the updated range for temperature.

TheR1D (approved):
Looks good to me, thanks.

PR #228, opened 2023-05-06
Add version option

### LINKED ISSUE #161: Check version of ShellGPT and warn user

melpeln:
Fixes #161

For the tests, I am not sure how to include the results (they passed tho!).


TheR1D:
Hi @melpeln thank you for interest and effort. Since development of ShellGPT was going rapidly, there were significant changes in the code base related to core features, which eventually cause merge conflicts.

PR #230, opened 2023-05-07
Add support for organization

AmShaegar13:
This adds support for setting organization. When API usage is payed by a different account you need to set its organization ID. You can set it via environment or config.

I didn't want to change how prompting for the API key works because this should be the easy way to use `sgpt`. Most users wont need an organization so this would just confuse them.

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -129,7 +129,7 @@ def main(
     if editor:
         prompt = get_edited_prompt()
 
-    client = OpenAIClient(cfg.get("OPENAI_API_HOST"), cfg.get("OPENAI_API_KEY"))
+    client = OpenAIClient(cfg.get("OPENAI_API_HOST"), cfg.get("OPENAI_API_KEY"), cfg.get("OPENAI_ORGANIZATION", False))
 
     role_class = DefaultRoles.get(shell, code) if not role else SystemRole.get(role)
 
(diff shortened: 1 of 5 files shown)


AmShaegar13:
I fixed unit tests. However, when running integration tests in get 429 because payment information are not setup in my organization yet.

AmShaegar13:
I managed to run and adjust the tests. Should be fine now.

TheR1D:
Hey @AmShaegar13 
ShellGPT will be using [LiteLLM](https://docs.litellm.ai/docs/providers/openai) to manage settings for specific LLM providers.  Using latest sgpt release it is possible to define organization using environment variables: `export OPENAI_ORGANIZATION=...`. Given that, I'll pass on this, but thanks for the effort! 🍰

AmShaegar13:
@TheR1D Came back here after switching back to the original version of shell_gpt. Sadly, what you say is not true for me. I still get 429 Too Many Requests because my organization is not taken into account. I need to add the following line:

`handlers/handler.py:14`
```
 additional_kwargs = {
     "timeout": int(cfg.get("REQUEST_TIMEOUT")),
     "api_key": cfg.get("OPENAI_API_KEY"),
+    "organization": cfg.get("OPENAI_ORGANIZATION"),
     "base_url": None if base_url == "default" else base_url,
```

PR #234, opened 2023-05-08
Helm Chart for k8s deployment

DmarshalTU:
Helm chart created with service account and clusterRoleBinding to admin-cluster to obtain permissions to run operations 
all over the cluster

created image from default docker file with some changes:

```bash
ENV OPENAI_API_KEY="" # for the env openAI key(later forward from secret through  deployment)
COPY ./utils/kubectl /usr/local/bin/kubectl # gives sgpt use Kubectl tool to exec the command in the cluster easily 

# ENTRYPOINT ["sgpt"]
CMD ["sleep", "infinity"] # prevent the pod from dying(not good solution need to fix in future)
```

example of calling the pod for action with kubectl : 

image

example from inside the pod:

image


TheR1D:
Hi @DmarshalTU 
Thank you for interest and effort. In my opinion ShellGPT is not often used in k8s cluster, it is more "personal" use case I would say. There for I'm closing this PR.

PR #242, opened 2023-05-12
Shell script

chinarjoshi:
Overrides command not found handler and implements 2 keybindings:
  1. **Control + Space**: Toggle the buffer between natural language and bash.
  2. **Enter**: If command not found, then if the command is natural language, translate and execute.
  3. **Control + Enter**: Execute the natural language in the buffer, regardless whether its also a valid bash command *("find all files longer than 20 lines")*. Also useful for Mac because of the uppercase commands.

One last feature to implement is more robust checking for destructive commands.

### DIFF
--- a/shell_gpt.sh
+++ b/shell_gpt.sh
@@ -142,32 +142,38 @@ _sgpt_override_enter_bash() {
 # If the command is not destructive or the user confirms, then return 0.
 # Otherwise, return 1.
 _sgpt_handle_destructive_command() {
-    local no_confirm_config_option='SHELL_EXECUTE_NO_CONFIRM'
-
-    # Get no-confirm configuration
-    local no_confirm
-    if [[ -f ~/.config/shell_gpt/.sgptrc ]]; then
-        no_confirm=$(grep -E '^${no_confirm_config_option}=' ~/.config/shell_gpt/.sgptrc \
-                | cut -d'=' -f2- | tr -d '[:space:]')
+    local config_path=~/.config/shell_gpt/.sgptrc
+    declare -A config_options=(
+        [no_confirm]='SHELL_EXECUTE_NO_CONFIRM'
+        [default_execute]='DEFAULT_EXECUTE_SHELL_CMD'
+    )
+    if [[ -f $config_path ]]; then
+        config_options[no_confirm]=$(. $config_path && eval echo \$$config_options[no_confirm])
+        config_options[default_execute]=$(. $config_path && eval echo \$$config_options[default_execute])
     fi
+
+    local confirm_string=$([[ $config_options[default_execute] == true ]] \
+                            && echo 'Y/n' || echo 'y/N')
 
     # If the command is destructive, then prompt for confirmation
     if _sgpt_is_destructive_command "$1"; then
         # Then prompt for confirmation
         echo -e "\033[1;31m${1}\033[0m" # Print the command in red
-        echo -e "Execute \033[1mdestructive\033[0m shell command? [y/N]:" # destructive in bold
+        echo -e "Execute \033[1mdestructive\033[0m shell command? [$confirm_string]:"
         read confirmation < /dev/tty
         # If user types enter or a word starting with n or N, then exit unsuccessfully
-        if [[ -z $confirmation || ${confirmation:0:1} == [nN] ]]; then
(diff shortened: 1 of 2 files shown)


chinarjoshi:
The shell script is now complete. Can someone please test it and check for any fixes?

chinarjoshi:
Something else to consider is putting this shell script in the main branch as a hidden file ".sgpt.sh", and on installation prompt them to opt in or out of modifying their shell rc to source it. I think that would allow more users to discover the feature compared to an install script from the README.

TheR1D (commented):
Thank you for the PR. I've noticed a few things during testing that we might want to address:
* I recommend removing the direct execution of shell commands without displaying the command to the user first. In this scenario, the user would have the responsibility of reviewing and executing commands, including potentially destructive ones.
* It might be beneficial to add a loading indicator during sgpt requests. Some commands could take a few seconds to load, for instance, Using docker create postgress container, forward ports and create todo database. During the request, the shell appears to "hang" until the process is completed, which isn't the best experience for the user.
* There are instances where the first word of a natural language input exists as a command, leading to confusing results. For example, with find all files with json extension in current folder, find is a valid command for the shell. When we toggle, it provides a description for the shell command, even though this was meant as a natural language request. Another example is install docker ('install' is a valid shell command for Mac users). A potential solution for these cases could involve using different key bindings for transforming natural language inputs into shell commands, and vice versa. However, I suspect there may be more efficient solutions available.

chinarjoshi:
### Summary of changes

1. New feature: Explicitly translate buffer from natural language to shell. Example use case: "ls in long form" -> "ls -l". Default keybinding: Alt+T, and can be overridden with new config option `TOGGLE_NATURAL_LANGUAGE_KEYBINDING`.
2. Executing natural language now always prompts by default, and gives a clear warning message for destructive commands. Users can execute non-destructive commands without confirmation using a new config option `SHELL_EXECUTE_NO_CONFIRM`.
3. Executing natural language respects the existing config option `DEFAULT_EXECUTE_SHELL_CMD`.
5. Adds a loading indicator, ⌛ , whenever sgpt is invoked internally.
6. Default keybindings are now Alt+t for toggling buffer and Alt+e for explicitly executing natural language. Enter remains the same. These two keybindings can be overridden with config options `TOGGLE_BUFFER_KEYBINDING` and `EXECUTE_NATURAL_LANGUAGE_KEYBINDING` respectively.

chinarjoshi:
1. Changed default keybindings from Alt+* to Control+*.
2. Changed `zle redisplay` invocations to `zle -I && zle redisplay`. Hopefully this combination is compatible for both Mac and Linux. I was a bit confused that `zle -I` fixes because according to its man page it prepares the widget for stdout, which is strange that changing the buffer counts as stdout on Mac.
3. I think the simplest solution is to put the shell script in the main branch as `/scripts/.sgpt.sh` or `/scripts/integrations/*.sh`, so it gets automatically updated with pip. Then, the only command the user has to run to opt-in to the feature is `echo '. ~/.local/lib/python3.11/site-packages/sgpt/... >> ~/.zshrc`. This would require adding some line to pyproject.toml to include the script as a data file. I found a way to do this with [poetry](https://python-poetry.org/docs/pyproject/#include-and-exclude) and [setuptools](https://setuptools.pypa.io/en/latest/userguide/datafiles.html), but not without either.

TheR1D:
If we move integration scripts to main and distribute it with `sgpt` we will be forced to make new pip package updates every time we would need to update shell integration scripts. In my opinion shell integration scripts should be completely separate with independed update logic. I think we can't just `echo some.sh >> ~/.zshrc`, if user already has shell integration written in shell profile, it will just keep adding new version at the end of file. My idea was to find some way to update our shell integraion lines in user shell profiles. Also it would be nice to keep bash/zsh version separate and deliver only lines which are related to user `$SHELL`.

chinarjoshi:
Important change: I have moved the shell script to a separate repository to decouple it from the Python codebase and address the installation issue. This decoupling allows the script to be used with any command-line GPT program, while keeping shell_gpt as the default program. I also made a [pr to ohmyzsh](https://github.com/ohmyzsh/ohmyzsh/pull/11779)  to streamline installation.

https://github.com/chinarjoshi/englizsh/

PR #244, opened 2023-05-13
added a function to log shell commands executed by sgpt

adilansari488:
## What is this PR for
As a system administrator, it is useful and required to know which commands are being executed in the system.
This PR adds the functionality to log shell commands executed by the sgpt. It only logs shell commands and does not log chat or code. 
It logs shell commands at the ***shell_gpt/shell_history*** location. 

## Test the added functionality
I have tested the added functionality and it is working as expected and not breaking any of the current functionality.

image

## Running scripts/lint.sh and scripts/test.sh
I have executed ***scripts/lint.sh*** and it gave the following output :

image

***scripts/test.sh*** is giving HTTPERROR 429 due to I am using free version of OpenAI API and it has nothing to do with the code changes.

### DIFF
--- a/sgpt/utils.py
+++ b/sgpt/utils.py
@@ -36,6 +36,20 @@ def get_edited_prompt() -> str:
         raise BadParameter("Couldn't get valid PROMPT from $EDITOR")
     return output
 
+def save_shell_history(command: str) -> None:
+    """
+    Creates a file in shell_gpt directory.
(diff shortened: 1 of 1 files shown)


TheR1D:
Thank you for the effort. This issue should be solved by https://github.com/TheR1D/shell_gpt/pull/267 which will put commands into user's default history.

Issue #185, opened 2023-04-16
When generating shell commands, please make `Y` the default action when pressing Enter


Boscop:
Thanks for making this, it's very useful, especially compared to all the other similar shell GPT scripts which don't pass the OS and $SHELL as context. I really appreciate this :)

I have a suggestion:
In most cases when using the `--shell`/`-s` flag, the user wants to actually run the generated command.
The only question is then, whether GPT got the command right.
In most cases it gets it right, so a single enter press to run it would be the most convenient user experience.
Please make `Y` the default for the `--shell` command.

E.g. instead of:
```sh
❯ sgpt -s "list all env vars"
Get-ChildItem Env:
Execute shell command? [y/N]: y
```
the user would only have to press Enter:
```sh
❯ sgpt -s "list all env vars"
Get-ChildItem Env:
Execute shell command? [Y/n]: 
```
and if the user doesn't want that, they can cancel with Ctrl-C (which would also be faster than entering `n` and pressing Enter).

xsoc:
Love this, even if it is just configurable via `~/.config/shell_gpt/.sgptrc`

PR #246, opened 2023-05-15
Create configuration parameter DEFAULT_EXECUTE_SHELL_CMD

konstantin-goldman:
This addresses the following issue: https://github.com/TheR1D/shell_gpt/issues/185

I wanted a mechanism to automatically execute GPT-suggested shell commands without entering "y" for each command. Others have asked for this, so I decided to open the PR.

### Testing:

Testing the case where the environment variable is true:
```
% DEFAULT_EXECUTE_SHELL_CMD=true ./env/bin/python ./sgpt/app.py -s "2 + 3"
echo $((2+3))
Execute shell command? [Y/n]:
5
% 
```

Testing the case where the environment variable is false (default):
```
% DEFAULT_EXECUTE_SHELL_CMD=false ./env/bin/python ./sgpt/app.py -s "2 + 3"
echo $((2+3))
Execute shell command? [y/N]:
%
```

Users can set the parameter in their config file `~/.config/shell_gpt/.sgptrc` or as an environment variable. The parameter defaults to the value "false", so the behavior after merging this PR is the same as the present behavior. I've added a note in the README about this new parameter, and will put a comment in the open issue if this is accepted.

Tested locally, both `./scripts/lint.sh` and `./scripts/tests.sh` passed successfully. I did not add an integration test because it seems non-trivial to spin-up a test session which dynamically interacts with the response from the GPT runner given the current test framework. Happy to take any feedback on that, perhaps there is a simpler way that I'm not aware of.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,8 @@ DEFAULT_MODEL=gpt-3.5-turbo
 DEFAULT_COLOR=magenta
 # Force use system role messages (not recommended).
 SYSTEM_ROLES=false
+# When in --shell mode, default to "Y" for no input.
+DEFAULT_EXECUTE_SHELL_CMD=false
 ```
 Possible options for `DEFAULT_COLOR`: black, red, green, yellow, blue, magenta, cyan, white, bright_black, bright_red, bright_green, bright_yellow, bright_blue, bright_magenta, bright_cyan, bright_white.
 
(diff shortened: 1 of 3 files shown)


TheR1D (approved):
Looks good to me, thank you.

PR #248, opened 2023-05-19
Add support for multiple words without quotes

Frodotus:
This pull request adds functionality to allow the use of prompts with multiple words without requiring the words to be enclosed in quotes. This enhancement improves the usability of command-line interface and reduces potential errors that can occur from forgetting to use quotes. Additionally, this pull request includes a default .gitignore file for Python, improving the developer experience by excluding irrelevant files from version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,36 @@
+Create a new file named ".gitignore" in the root directory of your Python app and add the following lines:
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# Local development settings
+.env
+
+# IDE files
+.vscode/
+.idea/
+
+# Dependency directories
+venv/
+env/
+env.bak/
+
+# Distribution / packaging
+dist/
+build/
+*.egg-info/
+*.egg
+
+# Compiled Python files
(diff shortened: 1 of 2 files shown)


TheR1D:
Thank you for the PR.
We already had simillar pull requests https://github.com/TheR1D/shell_gpt/pull/71. From default shell behaviour perspective it feels a bit unusual personally for me. But I believe many `sgpt` users at this point could benefit from this feature. I would suggest to make this as optional feature using [config](https://github.com/TheR1D/shell_gpt/blob/main/sgpt/config.py) parameter.

TheR1D (commented):
Thank you for the PR.
We already had simillar pull requests https://github.com/TheR1D/shell_gpt/pull/71. From default shell behaviour perspective it feels a bit unusual personally for me. But I believe many `sgpt` users at this point could benefit from this feature. I would suggest to make this as optional feature using [config](https://github.com/TheR1D/shell_gpt/blob/main/sgpt/config.py) parameter.

TheR1D:
It is solved in https://github.com/TheR1D/shell_gpt/releases/tag/0.9.2 by shell integrations.

Issue #256, opened 2023-05-29
CACHE_PATH and CHAT_CACHE_PATH var issues in a multiuser system.


shellclear:
When different users run the program the default value of the CACHE_PATH and CHAT_CACHE_PATH var appoint to the same folder inside _/tmp /cache_ directory generating perms issues.

The program should create a cache folder for each user inside /tmp directory.

TheR1D:
These changes will cause all current users to lose their chat sessions and cache. May be safe solution in this case would be to edit `CHAT_CACHE_PATH` and `CACHE_PATH` in `.sgptrc` per user manually.

PR #257, opened 2023-05-29
Adding uid to the name of CACHE_PATH and CHAT_CACHE_PATH vars

shellclear:
Hi, I propose this change to solve the permission issues when using shell-gpt in a multiuser system.


shellclear:
https://github.com/TheR1D/shell_gpt/issues/256

PR #259, opened 2023-05-30
using uid to define CACHE_PATH and CHAT_CACHE_PATH var

shellclear:
Hi, I propose this change to solve the permission issues when using shell-gpt in a multiuser system.

### DIFF
--- a/sgpt/config.py
+++ b/sgpt/config.py
@@ -12,8 +12,8 @@ CONFIG_FOLDER = os.path.expanduser("~/.config")
 SHELL_GPT_CONFIG_FOLDER = Path(CONFIG_FOLDER) / "shell_gpt"
 SHELL_GPT_CONFIG_PATH = SHELL_GPT_CONFIG_FOLDER / ".sgptrc"
 ROLE_STORAGE_PATH = SHELL_GPT_CONFIG_FOLDER / "roles"
-CHAT_CACHE_PATH = Path(gettempdir()) / "chat_cache"
-CACHE_PATH = Path(gettempdir()) / "cache"
+CHAT_CACHE_PATH = Path(gettempdir()) / f"sgpt-chat-cache-{os.getuid()}"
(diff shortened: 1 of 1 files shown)


shellclear:
#256

sgpt/config.py
```diff
@@ -12,8 +12,8 @@
 SHELL_GPT_CONFIG_FOLDER = Path(CONFIG_FOLDER) / "shell_gpt"
 SHELL_GPT_CONFIG_PATH = SHELL_GPT_CONFIG_FOLDER / ".sgptrc"
 ROLE_STORAGE_PATH = SHELL_GPT_CONFIG_FOLDER / "roles"
-CHAT_CACHE_PATH = Path(gettempdir()) / "chat_cache"
-CACHE_PATH = Path(gettempdir()) / "cache"
+CHAT_CACHE_PATH = Path(gettempdir()) / f"sgpt-chat-cache-{os.getuid()}"
+CACHE_PATH = Path(gettempdir()) / f"sgpt-cache-{os.getuid()}"
```
TheR1D, line 16: This change might cause issues for Windows users with Shell-GPT because os.getuid is not supported on their platform. A potential solution could be to use the [getuser](https://docs.python.org/3/library/getpass.html#getpass.getuser) function instead, which is more cross-platform friendly.

TheR1D (changes requested):

sgpt/config.py
TheR1D: Also these changes will cause all users to lose their chat sessions and cache. May be safe solution in this case would be to edit `CHAT_CACHE_PATH` and `CACHE_PATH` in `.sgptrc` per user manually.

TheR1D:
Thank you for the Pull Request. These changes will cause all users to lose their chat sessions and cache. May be safe solution in this case would be to edit `CHAT_CACHE_PATH` and `CACHE_PATH` in `.sgptrc` per user manually.

PR #263, opened 2023-06-02
Model endpoint compatibility

will-wright-eng:
## Summary

Adding models found in [OpenAI docs for model compatibility](https://platform.openai.com/docs/models/model-endpoint-compatibility). I found that the responses have started to drift in their length, so I'd like to set the frozen version of `gpt-3.5-turbo` as my default (`gpt-3.5-turbo-0301`)

### Setup

```bash
conda create -n test python=3.10 -y
conda activate test
python -m pip install -e ."[dev,test]"
python -m pytest
```

### Results

```bash
(test) ➜  shell_gpt git:(adding-models) ✗ python -m pytest                                     0:03:08
========================================= test session starts =========================================
platform darwin -- Python 3.10.11, pytest-7.3.1, pluggy-1.0.0
rootdir: /Users/will-wright-eng/shell_gpt
plugins: requests-mock-1.10.0
collected 24 items

tests/test_integration.py ......................                                                [ 91%]
tests/test_unit.py ..                                                                           [100%]

==================================== 24 passed in 87.27s (0:01:27) ====================================
```

### DIFF
--- a/sgpt/utils.py
+++ b/sgpt/utils.py
@@ -10,9 +10,16 @@ from click import BadParameter
 
 
 class ModelOptions(str, Enum):
-    GPT3 = "gpt-3.5-turbo"
-    GPT4 = "gpt-4"
-    GPT4_32K = "gpt-4-32k"
+    """
+    Model endpoint compatibility
+    https://platform.openai.com/docs/models/model-endpoint-compatibility
+    """
+    GPT4 = 'gpt-4'
+    GPT40314 = 'gpt-4-0314'
+    GPT432k = 'gpt-4-32k'
+    GPT432k0314 = 'gpt-4-32k-0314'
+    GPT35TURBO = 'gpt-3.5-turbo'
+    GPT35TURBO0301 = 'gpt-3.5-turbo-0301'
 
 
 def get_edited_prompt() -> str:
--- a/sgpt/utils.py
+++ b/sgpt/utils.py
@@ -14,17 +14,18 @@ class ModelOptions(str, Enum):
     Model endpoint compatibility
     https://platform.openai.com/docs/models/model-endpoint-compatibility
     """
-    GPT4 = 'gpt-4'
-    GPT40314 = 'gpt-4-0314' # discontinued 2023-09-13
-    GPT40613 = 'gpt-4-0613'
-    GPT432k = 'gpt-4-32k'
-    GPT432k0314 = 'gpt-4-32k-0314' # discontinued 2023-09-13
-    GPT432k0613 = 'gpt-4-32k-0613'
-    GPT35TURBO = 'gpt-3.5-turbo'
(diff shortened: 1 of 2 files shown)


mountaineerbr:
Nice observation! Indeed, I find that OpenAI models are getting CENSORED, and their replies always more politically-correctly which is just...

There seems to be some fun left with the original "davinci" and "curie" models, though (these are trained with internet data)!

Anyways, it does not matter if your request is set for `gpt-3.5-turbo`, because all answers that come back are from `gpt-3.5-turbo-0301` (check the JSON response). This option will be useful only when there are other versions of `gpt-3.5-turbo` available, I reckon..

will-wright-eng:
"lobotomized parrot" 🤣

Good catch on the model response -- I misinterpreted the documentation. Found [a community post](https://community.openai.com/t/attempting-to-use-gpt-3-5-turbo-getting-gpt-3-5-turbo-0301-instead/94411) discussing exactly this. In that case, this PR is less relevant until a new 3.5 model is released or if you have bypassed the waitlist for gpt-4 API access

sgpt/utils.py
```diff
@@ -10,9 +10,16 @@
...
+    """
+    Model endpoint compatibility
+    https://platform.openai.com/docs/models/model-endpoint-compatibility
+    """
+    GPT4 = 'gpt-4'
+    GPT40314 = 'gpt-4-0314'
+    GPT432k = 'gpt-4-32k'
+    GPT432k0314 = 'gpt-4-32k-0314'
+    GPT35TURBO = 'gpt-3.5-turbo'
+    GPT35TURBO0301 = 'gpt-3.5-turbo-0301'
```
TheR1D, line 22: I have concerns about including models that will be phased out with each new version. The OpenAI documentation specifically states:

> This model will not receive updates and is scheduled for deprecation 3 months post the release of a new version.

It may not be efficient to utilize such models, given their short lifespan and lack of continued support.

will-wright-eng: I think this is a valid concern -- it feels like needless maintenance to constantly be updating a dataclass. However, without the update it limits user's ability to choose the model they are using. With [today's announcement](https://openai.com/blog/function-calling-and-other-api-updates) this includes a range of model options...

### I see three options:
1. only allow for the default model option: by only updating `ModelOptions` with non-date models that wont be deprecated
2. update with all available options: (as above) feels like needless maintenance
3. remove the data class: least restrictive but will return ambiguous error when a non-compatible model is used

**Personally I think 2 is low-effort with respect to PR review time, and adds zero complexity to the control flow**

Your call!

will-wright-eng: *Updated PR with all options*

TheR1D: I personally prefer the first option. However, for the time being, let's proceed with merging this. I'll ensure to remove any non-date models prior to releasing the next version of sgpt.

TheR1D (approved):

PR #266, opened 2023-06-04
repl changes

DeviousD4n:
Simple repl changes to make the flag shorter and ability to enter request at the same time. Forgive me if there are any issues i've overlooked as i'm new to github and programming!

### DIFF
--- a/sgpt/handlers/repl_handler.py
+++ b/sgpt/handlers/repl_handler.py
@@ -30,6 +30,9 @@ class ReplHandler(ChatHandler):
         )
         typer.secho(info_message, fg="yellow")
 
+        if prompt:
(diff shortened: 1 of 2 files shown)


sgpt/handlers/repl_handler.py
```diff
@@ -30,6 +30,9 @@ def handle(self, prompt: str, **kwargs: Any) -> None:  # type: ignore
         )
         typer.secho(info_message, fg="yellow")
 
+        if prompt:
+           super().handle(prompt, **kwargs)
```
TheR1D, line 34: In this case, if we strat REPL with `--shell` the first command suggestion will not be possible to execute.

PR #272, opened 2023-06-07
added a check for empty var

kyrbrbik:
Shell integrations behaves weird if the command line is empty. 
Added a basic check for bash to not execute with empty command line.

### DIFF
--- a/simple_bash.sh
+++ b/simple_bash.sh
@@ -1,7 +1,9 @@
 # Shell-GPT integration BASH v0.1
 _sgpt_bash() {
-    READLINE_LINE=$(sgpt --shell <<< "$READLINE_LINE")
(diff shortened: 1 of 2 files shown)


simple_bash.sh
```diff
@@ -1,7 +1,9 @@
 # Shell-GPT integration BASH v0.1
 _sgpt_bash() {
-    READLINE_LINE=$(sgpt --shell <<< "$READLINE_LINE")
+if [[ -n "$READLINE_LINE" ]]; then
```
TheR1D, line 3: Could you please add same check for zsh version?

kyrbrbik:
This should do it

TheR1D (approved):
LGTM

PR #280, opened 2023-06-14
Fix README fizzbuzz: Modify from python -m sgpt to just sgpt

startakovsky:


TheR1D (approved):
Thank you, looks good to me.

Issue #289, opened 2023-06-17
Add "disable streaming" option


danarth:
In #172 there is a short discussion around disabling the "typewriter" animation.

I have a similar requirement to query the OpenAI API without streaming, so I would like to work on a new config property that allows streaming to be disabled

PR #290, opened 2023-06-17
Add option to disable streaming #289

danarth:


TheR1D (commented):
Could we consider incorporating a progress indicator while waiting for the complete response? A basic "loading..." message displayed in the standard output could suffice, in my opinion.

danarth:
Hi @TheR1D , thanks for the feedback. I am happy to implement this, and I think `typer` has some nice progress indicators, but my only concern is that it changes the output of the command. For example, if someone were to pipe the output of the `sgpt` command, would we get “Loading…” in that output?

TheR1D:
> Hi @TheR1D , thanks for the feedback. I am happy to implement this, and I think `typer` has some nice progress indicators, but my only concern is that it changes the output of the command. For example, if someone were to pipe the output of the `sgpt` command, would we get “Loading…” in that output?

The answer is no, as we would replace the "Loading..." line with the desired output. While our application is running, we can manipulate the stdout stream as needed. Here is a simple illustration:
```python
import time

print("loading...", end="\r")
time.sleep(2)  # simulate waiting for data
print("Here is the actual data.")
```
The technique involves the use of the carriage return character (\r). This character when printed returns the print head back to the start of the line. So, any subsequent print will overwrite the existing content of the line. However, Typer allows us to execute this in a more elegant manner.

danarth:
@TheR1D I have added a loading message now - thanks!

TheR1D (approved):
Thank you, I added few minor changes.

PR #292, opened 2023-06-19
Moving shell integration scripts to main branch

TheR1D:


yarikoptic:
I thought I even filed an issue for this, so was pleasantly surprised to see the PR!

I might only suggest that now with such a move you can help people to avoid downloading it etc, but rather can ship them under `sgpt/` within your `sgpt` python package, and just access from there. Proses: 
- "vetted" version would be installed instead of downloading a possibly compromised script from the web
- you can craft `shell-completion` command alike we have e.g. in [datalad](https://github.com/datalad/datalad/blob/d119a09ccedb81d49ee0b6cb384e2457953089b9/datalad/interface/shell_completion.py#L38) and give users ability to just do `source <(sgpt shell-completion)` whenever they seek for selectively enabling it in the sessions where they needed it.

TheR1D:
Resolved in https://github.com/TheR1D/shell_gpt/pull/455

PR #343, opened 2023-09-27
Create python-package-conda.yml

13aalteshuva:


TheR1D:
Hi @13aalteshuva thank you for interest and effort. ShellGPT originally designed using pip package manger, I think it will make it a bit complicated if we add another package/env management tool. Therefore I'm closing this PR, but thank you for your effort.

PR #349, opened 2023-10-04
On shell mode - Adding Edi[t] and [C]opy options

udiNur:
Iv'e  added options on shell command mode:
1. Edit the provided command before runnnig
2. Copy the provided command to clipboard 

If there is anything that missing let me know

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -3,6 +3,8 @@ import readline  # noqa: F401
 import sys
 
 import typer
+import pyperclip
+import readline
 from click import BadArgumentUsage, MissingParameter
 from click.types import Choice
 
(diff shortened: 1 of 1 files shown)


levid0s (changes requested):
You're introducing a new unnecessary dependency `pyperclip`.
I tested on Ubuntu connected via putty, and neither Edit nor Copy worked.

Copy error:
```
PyperclipException:                                                                                               
    Pyperclip could not find a copy/paste mechanism for your system.                                              
    For more information, please visit https://pyperclip.readthedocs.io/en/latest/index.html#not-implemented-error
```

With Edit, nothing happens:
```
lev@hppl:/tmp/shell_gpt$ python -m sgpt --shell "list disks"
lsblk                                                       
[E]xecute, Edi[t], [C]opy, [D]escribe, [A]bort: t           
lev@hppl:/tmp/shell_gpt$                                    
```

The Edi[t] menu item choice with the **t** shortcut is not great, because I inadvertently selected **E** thinking it's edit but it's actually execute! Probably Execute should be X and Edit should be E or T.

vansoest:
Good idea—I would also suggest add new sub arguments to the --shell argument for, e.g., edit directly the shell command without pressing the action key. This would be handy if you define it as a shell alias, to output and edit fast shell commands.

TheR1D:
Hi @udiNur
Thank you for interest and effort. This problem (copy/edit) is solved by [shell integration](https://github.com/TheR1D/shell_gpt#shell-integration) feature, which basically allows to do copy/edit natively. Therefore I'm closing this PR, but thank you for your effort.

PR #355, opened 2023-10-19
Configurable default Temp and TopP

jordac:


TheR1D:
Hi @jordac 
Thank you for interest and effort. I think tools like ShellGPT benefits from having low and constant tempreature, since it makes LLM response more precise. Therefore I'm closing this PR, thank you for your effort.

jordac:
https://www.reddit.com/r/ChatGPT/comments/126sr15/gpt_api_analyzing_which_temperature_and_top_p/

This post and my own testing is why I set both of mine to 0.3. Thanks for looking over my PR and for authoring this project,

image

jordac:
Having the higher top_p will cause results to be less precise.

```
The top_p parameter, often used in language models like GPT-3 and GPT-4, refers to "nucleus sampling," a method for generating text. It's a way to balance the randomness and predictability of the text being generated.

top_p = 1.0: This setting means that the model considers the full range of possible next words, as long as their cumulative probability reaches 100%. Essentially, it's looking at a wider range of words, including both high and low probability options. This can lead to more diverse and creative outputs but can also introduce more randomness and potentially less precise or relevant answers.

top_p = 0.3: This setting is more restrictive. The model only considers the smallest set of words whose cumulative probability adds up to 30%. It's focusing on the most likely next words and excludes the less probable ones. This generally leads to more predictable, coherent, and precise outputs but can sometimes be less creative or diverse in its responses.

In summary, a lower top_p (like 0.3) typically results in more precise and coherent text, as it's relying on the most probable words. A higher top_p (like 1.0) allows for more diverse and potentially creative outputs but can also introduce more randomness and less precision. The choice between the two depends on whether the priority is creativity and variety or precision and relevance.
```

PR #359, opened 2023-10-31
Make optional HTTP-referer configurable

ecoughlan:
This makes the HTTP Referrer configurable, which is required for requests to make `OPENAI_API_HOST=https://openrouter.ai/api` work (and then together with #358 it just works :tm:, with all their models). Spelling is consistent with header name, but could also be "proper" instead.

This implementation is a bit awkward and uses a special string "<none>" for unset, because Config.get() requires truthy values, so it can't just be left empty. But at least this way it's very unlikely to break anything for anyone.

### DIFF
--- a/sgpt/client.py
+++ b/sgpt/client.py
@@ -16,9 +16,10 @@ DISABLE_STREAMING = str(cfg.get("DISABLE_STREAMING"))
 class OpenAIClient:
     cache = Cache(CACHE_LENGTH, CACHE_PATH)
 
-    def __init__(self, api_host: str, api_key: str) -> None:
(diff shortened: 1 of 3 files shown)


TheR1D:
Hi @ecoughlan
Thank you for interest and effort. ShellGPT originally was designed to be used in pair with OpenAI API (things like prompts, roles, etc...). While using it with `https://openrouter.ai/api` (which also provides a lot of other LLMs) more likely a "personal" usecase. Also future migration to [assistant API](https://platform.openai.com/docs/assistants/overview) might break such multiple API integrations. Therefore I'm closing this PR, thank you for you effort.

PR #360, opened 2023-10-31
Azure OpenAI Service

MikeCer:
Hi team, I've added the support to Azure OpenAI services by simply introducing a new client class (that now is essentially the same, but with different endpoint and auth header). The handler will use the right class by evaluating a new param in the cfg kv collection.


TheR1D:
Hey @MikeCer thank you for effort and interest. Since there are a lot of LLM providers, ShellGPT will be using [LiteLLM](https://github.com/BerriAI/litellm) to cover multiple LLM providers. It was implemented in https://github.com/TheR1D/shell_gpt/pull/463 and released as 1.3.0. Given that, I'll pass on this, but thanks for the effort! 🍰

Issue #333, opened 2023-08-21
Render markdown output


volkov:
Currently output of command is not formated by default
![image](https://github.com/TheR1D/shell_gpt/assets/459989/c24e82f1-d096-4b90-87a7-5bbbd2a303a7)

It would be more readable if use something like `glow` formatting (although glow is written in go - so some alternative should be found)
![image](https://github.com/TheR1D/shell_gpt/assets/459989/241eca61-1635-4d0b-ab9d-980ae988c94b)

TheR1D:
We can pipe output of `sgpt` to `glow` or any other markdown reader.
```shell
sgpt "how to zip all files in current folder" | glow
```

volkov:
True, at least for latest glow versions (1.4.0 does not support stdin).

David-Else:
How about https://github.com/swsnr/mdcat?

jeanlucthumm:
https://rich.readthedocs.io/en/stable/markdown.html

^ python lib for markdown rendering with code syntax highlighting

jeanlucthumm:
The line that prints messages seems to be this one:

https://github.com/TheR1D/shell_gpt/blob/1c585664889e2cfb19ca1e13f0e8d621463ddef3/sgpt/handlers/handler.py#L34

Gonna try to change it to the Rich markdown handler and see what happens

jeanlucthumm:
To deal with the streaming, one option is to detect beginning of special blocks, like a code block, and then automatically append a closing statement until we get the actual one from the llm. We'd have to identify all such blocks.

Another consideration is how slow the markdown render is. It has to fully complete for every word streamed in which is multiple times per second.

jeanlucthumm:
I found the `DISABLE_STREAMING` option and at the very least, it's possible to output markdown if that's enabled:



https://github.com/TheR1D/shell_gpt/assets/4934853/d715bad9-01c6-4618-aa26-9607b76bafb7



Code: https://github.com/jeanlucthumm/shell_gpt/commit/eed0f7e813ab61a3c653ef2b03b34a4329ddd6ee

I'm not sure how I feel about the long waiting times though.

Other than that, I think you would need a full screen TUI to pull this off because you have to clear and repain the markdown file as words come in, but only from the current line down. The only thing I found was clearning the entire screen using `console.clear()`, and while that works, you lose all your history.

gwpl:
[...]

PR #377, opened 2023-11-22
add syntax highlighting via themeing

jaycenhorton:
Related to #333, this approach enables syntax highlighting, including for streaming chats

sgpt-theme

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,6 +347,8 @@ SYSTEM_ROLES=false
 DEFAULT_EXECUTE_SHELL_CMD=false
 # Disable streaming of responses
 DISABLE_STREAMING=false
+# The pygment theme to use for syntax highlighting
+CODE_THEME=one-dark
 ```
 Possible options for `DEFAULT_COLOR`: black, red, green, yellow, blue, magenta, cyan, white, bright_black, bright_red, bright_green, bright_yellow, bright_blue, bright_magenta, bright_cyan, bright_white.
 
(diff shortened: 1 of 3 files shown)


TheR1D:
Thank you for the PR @jaycenhorton, looks good! The only thing I don't like about rich markdown render is [vertical overflow](https://rich.readthedocs.io/en/stable/live.html#vertical-overflow), when the output text is larger then terminal height, it renders elipsis at last line "...", until we will get full response from API. This is especially noticeable when you have small terminal window (not full screen). For regular user it might be confusing, since it looks like `sgpt` "hangs". 

But this might be usefull for many `sgpt` users. Can we make it as optional feature? Using `.sgptrc` we could define something like `RENDER_MARKDOWN=true` (false by default).

jeanlucthumm:
Regarding the vertical overflow, in the demo it seems like the terminal just scrolls when it hits the bottom. How is that achieved?

TheR1D:
> Regarding the vertical overflow, in the demo it seems like the terminal just scrolls when it hits the bottom. How is that achieved?

As I described in my previous comment, when having a small terminal window (not full screen), and requesting some long text from LLM (which can't fit in current terminal window height), we will have vertical overflow. 

Another thing which I was thinking about it that, by default ShellGPT needs plain text in order to exectue shell comamnds or create valid source code files. So this feature probably will require some tweaks in the default role, which will explicitly require markdown as output from LLM.

TheR1D:
Implemented in https://github.com/TheR1D/shell_gpt/pull/400 with few role changes. Included @jaycenhorton as Co-Author of the commit, thank you for your interest and effort.

PR #393, opened 2023-12-11
Add Multiline REPL feature

### LINKED ISSUE #369: Multi-line syntax for REPL [Feature request]
Two ideas (maybe both?) for Multi-line syntax for REPL [Feature request] : 

* like  aider

```
{
multi
line
}
```

* like bash

```
<<<MARKER_OF_CHOICE
multi
line
MARKER_OF_CHOICE
```

arafatsyed:
Allow users to insert multiple lines to a REPL input, useful for copypasting large inputs and better usability. 

For multiline inputs, press '"""' and enter your prompt and exit the prompt with '"""'.

image

Resolves #369


arafatsyed:
image

sgpt/handlers/repl_handler.py
```diff
@@ -34,6 +48,8 @@ def handle(self, prompt: str, **kwargs: Any) -> None:  # type: ignore
         while True:
             # Infinite loop until user exits with Ctrl+C.
             prompt = typer.prompt(">>>", prompt_suffix=" ")
+            if prompt == "{":
```
TheR1D, line 51: I think we can use Python like multi-line text syntax with triple quotes `"""` (which is well known):
```python
>>> """
... test
... text
... """
-> 'test\ntext'
```
So in this case `"""` triggers multi-line input (start/end) and `...` indicates we are currently in multi-line mode.

sgpt/handlers/repl_handler.py
```diff
@@ -14,6 +14,20 @@ class ReplHandler(ChatHandler):
     def __init__(self, chat_id: str, role: SystemRole) -> None:
         super().__init__(chat_id, role)
 
+    def get_multiline_input(self) -> str:
+        info_message = (
+            "Entering REPL multiline mode, press '}' to confirm and end multiline prompt."
```
TheR1D, line 19: I think this message is not necessary once we will have `"""` as a trigger.

TheR1D (changes requested):

sgpt/handlers/repl_handler.py
```diff
@@ -14,6 +14,15 @@ class ReplHandler(ChatHandler):
     def __init__(self, chat_id: str, role: SystemRole) -> None:
         super().__init__(chat_id, role)
 
+    def get_multiline_input(self) -> str:
```
TheR1D, line 17: This could be shorter using walrus operator, but not an issue.
```python
def get_multiline_input(self) -> str:
    multiline_input = ""
    while (user_input := typer.prompt("...", prompt_suffix="")) != '"""':
        multiline_input += user_input + "\n"
    return multiline_input
```

TheR1D (approved):

TheR1D:
Thank you @arafatsyed, merged.

PR #394, opened 2023-12-12
Add option the check version of sgpt

### LINKED ISSUE #161: Check version of ShellGPT and warn user

Navidur1:
Add an option to check the version of `sgpt` using `--version` or `-v`.

Resolves #161 

image


sgpt/version.py
TheR1D, line 1: Let's rename this file as `__version__.py`.

TheR1D (changes requested):

Navidur1:
@TheR1D fixed the previous suggestion! Let me know if there's anything else that needs to be fixed

pyproject.toml
```diff
@@ -60,7 +60,7 @@ dev = [
 ]
 
 [tool.hatch.version]
-path = "sgpt/__init__.py"
+path = "sgpt/version.py"
```
TheR1D, line 63: Need to change the file name here as well.

TheR1D (changes requested):

TheR1D (approved):

TheR1D:
Thanks, merged.

PR #395, opened 2023-12-12
Created --reset-key flag to update OpenAI key

Ismail-Ben:
Resolves: #30 

Added --reset-key flag to update OPENAI_API_KEY in .sgptrc file. Also verified with a dummy chat completion that the newly inputted key works.

Sample output with 2 non-valid keys followed by a real key:
image

### DIFF
--- a/sgpt/client.py
+++ b/sgpt/client.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from typing import Dict, Generator, List
 
 import requests
+import typer
 
 from .cache import Cache
 from .config import cfg
@@ -58,7 +59,15 @@ class OpenAIClient:
             timeout=REQUEST_TIMEOUT,
             stream=stream,
         )
-        response.raise_for_status()
+        try:
+            response.raise_for_status()
+        except requests.exceptions.HTTPError as http_err:
+            # Check if the status code is 403 (Forbidden) or 401 (Unauthorized)
+            if http_err.response.status_code == 403 or http_err.response.status_code == 401:
+                typer.secho("Invalid OpenAI API key.", fg="red")
+            else:
+                typer.secho(f"HTTPError occurred: {http_err}", fg="red")
+            raise typer.Exit(code=1)
         # TODO: Optimise.
         # https://github.com/openai/openai-python/blob/237448dc072a2c062698da3f9f512fae38300c1c/openai/api_requestor.py#L98
         if not stream:
--- a/sgpt/client.py
+++ b/sgpt/client.py
@@ -6,7 +6,7 @@ import requests
 import typer
 
 from .cache import Cache
-from .config import cfg
+from .config import cfg, SHELL_GPT_CONFIG_PATH
 
 CACHE_LENGTH = int(cfg.get("CACHE_LENGTH"))
 CACHE_PATH = Path(cfg.get("CACHE_PATH"))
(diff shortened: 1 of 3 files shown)


Ismail-Ben:
@TheR1D Any suggested changes to get this in?

TheR1D (changes requested):
I think this option doesn't align with prioritized environment variables in config now, since we can't reset user exported `OPENAI_API_KEY`. But we could add better error handling into `client._request` simply checking respone status code for `403/401` printing the error text and raising `Exit` with error `raise typer.Exit(code=1)`.

Ismail-Ben:
> I think this option doesn't align with prioritized environment variables in config now, since we can't reset user exported `OPENAI_API_KEY`. But we could add better error handling into `client._request` simply checking respone status code for `403/401` printing the error text and raising `Exit` with error `raise typer.Exit(code=1)`.

I've moved the error handling to client._request as suggested. 

As for the reset functionality I added a note to the "help" section of the command as a reminder that the environment variable will still be prioritized. I still believe having a --reset-key option is worthwhile as I would've liked to use it myself when first trying the tool. Instead of having to navigate to the .sgptrc file and update it manually I would've been nice to have this command. 

The config.get() function properly prioritizes the use of the env key if it exists so updating the .sgptrc file wouldn't change this behavior.  In the case that a user runs this command while already having the env variable I highlight in yellow that their env key will still be used.

User updates key with env variable defined:
image

No env variable defined, and fake key:
image

Help section:
image

Let me know what you think @TheR1D . Thanks!

sgpt/client.py
```diff
@@ -58,7 +59,15 @@ def _request(
             timeout=REQUEST_TIMEOUT,
             stream=stream,
         )
-        response.raise_for_status()
+        try:
```
TheR1D, line 62: There is no reason to use `try` block here, we can just check `response.status` above.

sgpt/client.py
```diff
@@ -58,7 +59,15 @@ def _request(
...
+            response.raise_for_status()
+        except requests.exceptions.HTTPError as http_err:
+            # Check if the status code is 403 (Forbidden) or 401 (Unauthorized)
+            if http_err.response.status_code == 403 or http_err.response.status_code == 401:
+                typer.secho("Invalid OpenAI API key.", fg="red")
```
TheR1D, line 67: We can add more info for user, something like `f"Invalid OpenAI API key, update your config file: {cfg.SHELL_GPT_CONFIG_PATH}"`.

sgpt/client.py
```diff
@@ -58,7 +59,15 @@ def _request(
...
+            else:
+                typer.secho(f"HTTPError occurred: {http_err}", fg="red")
```
TheR1D, line 69: We don't need this line since `raise_for_status` should handle it.

TheR1D (changes requested):
It is very common in CLI applications to use runtime configuration file or ENV variables to set API keys (without having dedicated CLI options to reset some of config values). Please remove reset key option, if we show user proper message, it is clear how to reset it.

Ismail-Ben:
Adjusted to reflect @TheR1D 's comments. When using an invalid key the output now looks like:
image

TheR1D (dismissed):
Thank you @Ismail-Ben

TheR1D:
Please check lint issues.

TheR1D:
Thank you, merged.

PR #398, opened 2023-12-19
System role optimisations for GPT4+ models

TheR1D:
* Optimising stadard ShellGPT roles.
* Forcing system roles when calling OpenAI API.
* Changing default model to GPT4 turbo.
* Minor code fixes and optimisations.


AlexanderYastrebov:
Hello, looks like there is no way to disable system role? Could be an interesting feature to talk to bare models.

PR #418, opened 2023-12-24
Added GPT-4 Vision Support & Token Counter

duysqubix:
Great project, not sure if its dead or not.. but with the new vision preview models. I have modified the original codebase to support vision by supplying a `--image` option. Only works with the `gpt-4-vision-preview` model, supports both local images and URLs.

Also added an extra dependency that estimates the number of tokens.


luandro:
Would be great to have vision available.

TheR1D:
Thank you for your contribution @duysqubix, and sorry for late reply. ShellGPT mostly was designed to work with text input, also it becomes difficult to support vision since some other non-OpenAI models don't have this feature. With that said I will close this PR for now.

PR #444, opened 2024-01-14
Fix #422: Markdown formatting for chat history

### LINKED ISSUE #422: Chat formatting when listing chat history
Default output is nicely formatted and I was expecting that printing chat history via `sgpt --show-chat 1` will print chat history with exactly same formatting but it's not the case. I see raw text without any format. Is there any option for formatting chat histories or this is just not implemented feature so far?

jeanlucthumm:
20240114_07h42m31s_grim

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,6 +365,8 @@ sgpt --role json_generator "random: user, password, email, address"
 }
 ```
 
+If the description of the role contains the words "APPLY MARKDOWN" (case sensitive), then chats will be displayed using markdown formatting.
+
 ### Request cache
 Control cache using `--cache` (default) and `--no-cache` options. This caching applies for all `sgpt` requests to OpenAI API:
 ```shell
--- a/sgpt/handlers/chat_handler.py
+++ b/sgpt/handlers/chat_handler.py
@@ -1,11 +1,11 @@
 import json
 from pathlib import Path
 from typing import Any, Callable, Dict, Generator, List, Optional
-from rich.console import Console
-from rich.markdown import Markdown
 
 import typer
 from click import BadArgumentUsage
+from rich.console import Console
+from rich.markdown import Markdown
 
 from ..config import cfg
 from ..role import DefaultRoles, SystemRole
--- a/sgpt/handlers/chat_handler.py
+++ b/sgpt/handlers/chat_handler.py
@@ -1,6 +1,8 @@
 import json
 from pathlib import Path
 from typing import Any, Callable, Dict, Generator, List, Optional
+from rich.console import Console
+from rich.markdown import Markdown
 
 import typer
 from click import BadArgumentUsage
@@ -108,14 +110,14 @@ class ChatHandler(Handler):
         return self.chat_session.exists(self.chat_id)
 
     @property
-    def initial_message(self) -> str:
-        chat_history = self.chat_session.get_messages(self.chat_id)
-        return chat_history[0] if chat_history else ""
-
-    @property
     def is_same_role(self) -> bool:
         # TODO: Should be optimized for REPL mode.
-        return self.role.same_role(self.initial_message)
+        return self.role.same_role(self.initial_message(self.chat_id))
+
+    @classmethod
+    def initial_message(cls, chat_id: str) -> str:
+        chat_history = cls.chat_session.get_messages(chat_id)
+        return chat_history[0] if chat_history else ""
 
     @classmethod
     @option_callback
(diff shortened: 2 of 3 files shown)


jeanlucthumm:
@TheR1D is there any procedure I need to do for assigning a reviewer?

TheR1D (changes requested):
Thanks @jeanlucthumm 
I think we need to apply Markdown rendering only for "default" role conversations. Since we can have `--code` conversations, and we are using pure text ouput there, markdown will skip `\n` and print multiple lines of code in one line.
```shell
sgpt --repl --code c1
# <- print hello and print world using Python
# -> print('hello')
# -> print('world')
# <- exit()
```
```shell
sgpt --show-chat c1
...
user: print hello and print world using Python                                                                                                   
assistant: print('hello') print('world') # <---
```

jeanlucthumm:
Ah ok good catch. Before enabling it only for `default`, what are your thoughts on wrapping assistant messages with markdown code block if we are in code mode? I think that would look very nice and it makes sense because the assistant is outputting code after all.

We can either change how the message is stored, or we can do this only for `show_messages()` by detecting mode and whether it's an assistant message.

TheR1D:
> Ah ok good catch. Before enabling it only for `default`, what are your thoughts on wrapping assistant messages with markdown code block if we are in code mode? I think that would look very nice and it makes sense because the assistant is outputting code after all.
> 
> We can do this only for `show_messages()` by detecting mode and whether it's an assistant message.
> 
> The main concern here would be not outputting markdown wrapper when user is expected to redirect to file. But I think the use case for that is only this:
> 
> ```
> sgpt --code --chat c1 "a python file that says hello" > hello.py
> sgpt --code "a python file that says hello" > hello.py
> ```
> 
> Both of which do not use `show_messages`

You're right, we can use markdown code block when showing `--code` conversations with `show_messages`. May be we can apply md formatting for `--shell` conversations as well. For instance wrap assistant responses with "inline code" using backticks ``, e.g. `ls -la | sort`

jeanlucthumm:
Sounds good. I'll need to change how conversations are stored for this because we don't currently store the role along side the messages, only a json list of messages, so `show-chat` has no way of knowing.

My idea is to change the format from a list of messages to an object like this:

```
{
  "role": "...",
  "messages": [
    ...
  ],
}
```

This is also good for future if we ever want to store more metadata about the conversations.

Will take a look this weekend.

jeanlucthumm:
Ok, @TheR1D please take a look.

Some demos:
code
default_role
shell

TheR1D:
Thank you, @jeanlucthumm. I will review it during this week. I believe we may need to make some changes.

TheR1D:
Hey @jeanlucthumm, thank you for the PR. After reviewing the changes, I believe this feature could be implemented more concisely. We can reuse the `initial_message` method available in `ChatHandler`. Here's an example:
```python
@classmethod
def initial_message(cls, chat_id: str) -> str:
    chat_history = cls.chat_session.get_messages(chat_id)
    return chat_history[0] if chat_history else ""
```
Then we can slightly change logic in `show_messages`.
```python
@classmethod
def show_messages(cls, chat_id: str) -> None:
    # Prints all messages from a specified chat ID to the console.
    init_message = cls.initial_message(chat_id)

    if "use and apply markdown" in init_message.lower():
        for message in cls.chat_session.get_messages(chat_id):
            if message.startswith("assistant"):
                Console().print(Markdown(message))
            else:
                typer.secho(message, fg=cfg.get("DEFAULT_COLOR"))
            typer.echo()
        return

    for index, message in enumerate(messages):
        color = "magenta" if index % 2 == 0 else "green"
        typer.secho(message, fg=color)
```
I think this is more suitable solution since we are following the system prompt, and it would be compatible with custom roles. It will not wrap `--code` conversations into a code block, but after giving it some more thought, I don't see many benefits in rendering the code using a markdown block. I think `--code` will mostly be used when a user wants to redirect clean code output elsewhere. If Markdown is required, we can prompt code generation using the default ShellGPT role. Could you also include tests to make sure it works as expected?

Thank you for interest and effort.

jeanlucthumm:
Sure, let's do it this way. Will take a look this weekend!

jeanlucthumm:
@TheR1D Ok I implemented your suggested change and also added unit tests. I also updated the documentation for custom roles so users are aware of this functionality.

Screenshot 2024-02-09 at 12 53 30 PM
Screenshot 2024-02-09 at 12 54 46 PM

sgpt/handlers/handler.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, role: SystemRole) -> None:
 
     @property
     def printer(self) -> Printer:
-        use_markdown = "APPLY MARKDOWN" in self.role.role
+        use_markdown = "apply markdown" in self.role.role.lower()
```
TheR1D, line 27: It was intentionally written in uppercase as `"APPLY MARKDOWN"` because we might encounter descriptions such as `"do not apply markdown"` In such cases, we will use markdown formatting for a role that, in fact, does not require markdown.

TheR1D (changes requested):
Just one minor comment, other than that, LGTM.

sgpt/handlers/handler.py
jeanlucthumm: Done.

jeanlucthumm:
Ok finished!

TheR1D (approved):

TheR1D:
Thank you @jeanlucthumm, merged.

PR #451, opened 2024-01-18
Add copy to clipboard feature

Mamut3D:
Add option to copy shell commands to clipboard in addtion to describe/execute option. Supports Termux terminal on android as well.


TheR1D:
Hi @Mamut3D, thank you for interest and effort. I believe we can achieve similar results with some shell commands, such as `sgpt -s --no-interaction "print hello world" | pbcopy` or by using [shell integrations](https://github.com/TheR1D/shell_gpt#shell-integration).

In an effort to maintain simplicity and clarity in our codebase, I suggest we utilize the capabilities of the shell environment
surrounding `sgpt` rather than adding similar features directly to the code. This approach helps us keep the project more maintainable and lightweight. I hope this makes sense.

Considering the information provided above, I'm closing this pull request. We can revisit the changes if needed in the future.

PR #470, opened 2024-02-09
Empty chunk choices

joouha:
Some OpenAI "compatible" APIs (e.g. https://lepton.ai) return a chunk with an empty list of choices at the end of their responses.

Currently this causes an error with shell-gpt as this case is not handled, which this PR fixes.


TheR1D:
This issue should be solved in the latest release by LiteLLM. Given that, I'll pass on this, but thanks for the effort! 🍰

PR #490, opened 2024-02-21
Add a [q] to quit option to REPL mode

heathdutton:
Since we already have [e] and [d] why not also a [q]?

Because we have "e" and "d", for some reason I keep assuming "q" is also an option, to which it responds something that makes me feel like a dummy every time.


devilcoder01:
@heathdutton Does this work in REPL mode?  
For example, if you want to type a word that starts with "q," does it quit the session or continue the conversation?

heathdutton:
Continue. The entire prompt must be "q" to quit.

TheR1D:
I think it is not nesesary to have some specifc promt in order quit REPL mode. I belive SIGINT (CTRL + C) is most common way to exit running apps on terminal.

With that said, I’ll go ahead and close this PR for now.

PR #504, opened 2024-03-06
Only split config lines on the first =

keiththomps:
If a config value (such as the API key) has an = in it then it will currently blow up when parsing the config file. This will ensure that config values are only split into 2 items, keeping the values intact.

My enterprise key has an `=` in it and this is what I see 
image


TheR1D:
@keiththomps thank you, merged.

PR #513, opened 2024-03-13
Fix --show-chat and --repl to respect --no-md

florian-lnx:
I noticed that the options `--show-chat` and `--repl` does no respect the `--no-md` option and just checks for the "APPLY MARKDOWN" text in the role description. I fixed that. All tests and linting passed.


TheR1D (approved):
Looks good to me, thank you @florian-lnx

PR #540, opened 2024-04-11
Container improvements

Robinsane:
1. added 3 env variables to make life easier configuring shell_gpt containers:
IN_CONTAINER: setting this one to true, helps fix some default things that aren't handy when in a container:
- The default for interaction becomes False, since [E]xecute on host doesn't work when you get the output from a container
- Shell_gpt will not check the machine for os and shell (which would be fetched from the container), but will check what the user has configured in following env vars: OS_OUTSIDE_CONTAINER and SHELL_OUTSIDE_CONTAINER

2. adjusted Dockerfile to use new env vars + added an example ollama Dockerfile


sgpt/config.py
```diff
@@ -34,6 +34,9 @@
     "API_BASE_URL": os.getenv("API_BASE_URL", "default"),
     "PRETTIFY_MARKDOWN": os.getenv("PRETTIFY_MARKDOWN", "true"),
     "USE_LITELLM": os.getenv("USE_LITELLM", "false"),
+    "IN_CONTAINER": os.getenv("IN_CONTAINER", "false"),
```
TheR1D, line 37: I think we should avoid using container specific ENV variables. Instead we can have only two additional vars: `OVERWRITE_SHELL_NAME` and `OVERWRITE_OS_NAME`.

Setting defaults to some specific OS/SHELL is confusing. I think we can set default values as `"default"`, later in the role definition, we can check if user wants to overwrite original OS/SHELL names.

sgpt/app.py
```diff
@@ -57,7 +57,7 @@ def main(
         rich_help_panel="Assistance Options",
     ),
     interaction: bool = typer.Option(
-        True,
+        True if cfg.get("IN_CONTAINER") == "false" else False,
```
TheR1D, line 60: We can have a new ENV varaible `SHELL_INTERACTION` for this case.

Dockerfile_ollama
TheR1D, line 1: Let's remove this Dockerfile. I seems we are assuming that user will manually install ollama into this container. I believe ShellGPT + Ollama setup should be done in some sort of docker-compose tool, using separate containers.

Dockerfile
```diff
@@ -1,5 +1,9 @@
 FROM python:3-slim
 
+ENV IN_CONTAINER=true
+ENV OS_OUTSIDE_CONTAINER="Linux/Red Hat Enterprise Linux 8.8 (Ootpa)"
```
TheR1D, line 4: This should be an argument, so user can setup the image with specific OS and SHELL names.

sgpt/role.py
```diff
@@ -62,7 +62,10 @@ def __init__(
     @classmethod
     def create_defaults(cls) -> None:
         cls.storage.parent.mkdir(parents=True, exist_ok=True)
-        variables = {"shell": cls._shell_name(), "os": cls._os_name()}
+        if cfg.get("IN_CONTAINER") != "true":
+            variables = {"shell": cls._shell_name(), "os": cls._os_name()}
```
TheR1D, line 66: I think it would be better to change `_os_name` and `_shell_name` functions, e.g.:
```python
@classmethod
def _shell_name(cls) -> str:
    overwrite_name = cfg.get("OVERWRITE_SHELL_NAME")
    if overwrite_name != "default":
        return overwrite_name

    current_platform = platform.system()
    if current_platform in ("Windows", "nt"):
        is_powershell = len(getenv("PSModulePath", "").split(pathsep)) >= 3
        return "powershell.exe" if is_powershell else "cmd.exe"
    return basename(getenv("SHELL", "/bin/sh"))
```

TheR1D (changes requested):
Could we also update Config and Docker section in README.md please?

ismaelchris:
These settings worked for me:

```
IN_CONTAINER=false
USE_LITELLM=true
```

sgpt/config.py
Robinsane: agreed
I find the name "default" a little confusing if the default is dynamic.
I'm opting for empty string as default value for now.

Dockerfile_ollama
Robinsane: 127.0.0.1 was a bad choice of me, I just quickly changed our production IP to not give up any info.
I just wanted to offer new users a quick way to set up like this.
I'll remove this file, but will mention the ENV vars in the README.md

Dockerfile
Robinsane: In the upcoming commit I'll leave it as ENV variable that is in the config.py DEFAULT_CONFIG.
Reason being that it might be handy to leave it interchangeable, e.g.:

- All of a sudden you need to work on a server with an os / shell you're very unfamiliar with. You can temporarily overwrite your os and shell name of your often used shellgpt to mock this environment, without having to install on that server / install in a similar virtual local environment
- Similar to above, if I'd like to use shellgpt containers for multiple setups: RHEL, Debian, Ubuntu, powershell, MacOs, ... I can run the same image each time, as long as I overwrite the OS and shell to mock these envs. If I had to make a different image each time that'd be a waste of diskspace and image build time in my opinion.

sgpt/config.py
TheR1D: We are using the `"default"` value for other config variables as well. Having empty string values in `.sgptrc` is confusing.

sgpt/app.py
```diff
@@ -57,7 +57,7 @@ def main(
         rich_help_panel="Assistance Options",
     ),
     interaction: bool = typer.Option(
-        True,
+        True if cfg.get("SHELL_INTERACTION") != "false" else False,
```
TheR1D, line 60:
```suggestion
        cfg.get("SHELL_INTERACTION") == "true",
```

sgpt/role.py
```diff
@@ -113,22 +113,28 @@ def get_role_name(cls, initial_message: str) -> Optional[str]:
...
+            return cfg.get("OVERWRITE_OS_NAME")
+        else:
+            current_platform = platform.system()
+            if current_platform == "Linux":
+                return "Linux/" + distro_name(pretty=True)
+            if current_platform == "Windows":
+                return "Windows " + platform.release()
+            if current_platform == "Darwin":
+                return "Darwin/MacOS " + platform.mac_ver()[0]
+            return current_platform
```
TheR1D, line 126:
```suggestion
        if cfg.get("OVERWRITE_OS_NAME") != "default":
            return cfg.get("OVERWRITE_OS_NAME")
        current_platform = platform.system()
        if current_platform == "Linux":
            return "Linux/" + distro_name(pretty=True)
        if current_platform == "Windows":
            return "Windows " + platform.release()
        if current_platform == "Darwin":
            return "Darwin/MacOS " + platform.mac_ver()[0]
        return current_platform
```

sgpt/role.py
```diff
@@ -113,22 +113,28 @@ def get_role_name(cls, initial_message: str) -> Optional[str]:
...
-            return "powershell.exe" if is_powershell else "cmd.exe"
-        return basename(getenv("SHELL", "/bin/sh"))
+        if cfg.get("OVERWRITE_SHELL_NAME") != "":
+            return cfg.get("OVERWRITE_SHELL_NAME")
+        else:
+            current_platform = platform.system()
+            if current_platform in ("Windows", "nt"):
+                is_powershell = len(getenv("PSModulePath", "").split(pathsep)) >= 3
+                return "powershell.exe" if is_powershell else "cmd.exe"
+            return basename(getenv("SHELL", "/bin/sh"))
```
TheR1D, line 137:
```suggestion
        if cfg.get("OVERWRITE_SHELL_NAME") != "default":
            return cfg.get("OVERWRITE_SHELL_NAME")
        current_platform = platform.system()
        if current_platform in ("Windows", "nt"):
            is_powershell = len(getenv("PSModulePath", "").split(pathsep)) >= 3
            return "powershell.exe" if is_powershell else "cmd.exe"
        return basename(getenv("SHELL", "/bin/sh"))
```

sgpt/config.py
```diff
@@ -34,6 +34,9 @@
     "API_BASE_URL": os.getenv("API_BASE_URL", "default"),
     "PRETTIFY_MARKDOWN": os.getenv("PRETTIFY_MARKDOWN", "true"),
     "USE_LITELLM": os.getenv("USE_LITELLM", "false"),
+    "SHELL_INTERACTION ": os.getenv("SHELL_INTERACTION ", "true"),
+    "OVERWRITE_OS_NAME": os.getenv("OVERWRITE_OS_NAME", ""),
+    "OVERWRITE_SHELL_NAME ": os.getenv("OVERWRITE_SHELL_NAME", ""),
```
TheR1D, line 39:
```suggestion
    "SHELL_INTERACTION ": os.getenv("SHELL_INTERACTION ", "true"),
    "OVERWRITE_OS_NAME": os.getenv("OVERWRITE_OS_NAME", "default"),
    "OVERWRITE_SHELL_NAME ": os.getenv("OVERWRITE_SHELL_NAME", "default"),
```

Dockerfile
```diff
@@ -1,5 +1,9 @@
 FROM python:3-slim
 
+ENV SHELL_INTERACTION=false
+ENV OVERWRITE_OS_NAME=""
+ENV OVERWRITE_SHELL_NAME=""
+
```
TheR1D, line 6:
```suggestion
ARG OVERWRITE_OS_NAME=default
ARG OVERWRITE_SHELL_NAME=default

ENV SHELL_INTERACTION=false
ENV OVERWRITE_OS_NAME=${OVERWRITE_OS_NAME}
ENV OVERWRITE_SHELL_NAME=${OVERWRITE_SHELL_NAME}

```

README.md
```diff
@@ -461,9 +461,16 @@ docker run --rm \
...
+* The \[E\]xecute option for --shell with interaction will not work, since it would try this Execute in the docker container.  
+=> setting the `SHELL_INTERACTION` environment variable to false , makes sense.
+* Since, most likely the os and shell of your container are not identical to the environment you want help with:  
+set the environment variables `OVERWRITE_OS_NAME` and `OVERWRITE_SHELL_NAME` according to your setup.
+
+
 Example of a conversation, using an alias and the `OPENAI_API_KEY` environment variable:
 ```shell
-alias sgpt="docker run --rm --env OPENAI_API_KEY --volume gpt-cache:/tmp/shell_gpt ghcr.io/ther1d/shell_gpt"
+alias sgpt="docker run --rm --env OPENAI_API_KEY --env SHELL_INTERACTION=false OVERWRITE_OS_NAME=Debian --env OVERWRITE_SHELL_NAME=bash --volume gpt-cache:/tmp/shell_gpt ghcr.io/ther1d/shell_gpt"
```
TheR1D, line 473:
```suggestion
alias sgpt="docker run --rm --volume gpt-cache:/tmp/shell_gpt --env OVERWRITE_OS_NAME=$(uname -s) --env OVERWRITE_SHELL_NAME=$(echo $SHELL) ghcr.io/ther1d/shell_gpt "
```

sgpt/app.py
Robinsane: I wrote it this way to ensure the normal default behaviour is set, also when SHELL_INTERACTION contains an unintended / weird value

sgpt/role.py
Robinsane: I find calling "not overwriting" the "overwriting"-env var  "default" a bit confusing. I prefer the empty string or maybe something like "no_overwrite". This is personal taste though, if you insist on "default" of course, go ahead.

The else was indeed redundant, my bad.

sgpt/role.py
Robinsane: I find calling "not overwriting" the "overwriting"-env var  "default" a bit confusing. I prefer the empty string or maybe something like "no_overwrite". This is personal taste though, if you insist on "default" of course, go ahead.

The else was indeed redundant, my bad.

sgpt/config.py
Robinsane: As stated above, I find the name "default" confusing for not using the var.
My personal preference goes to something like "", "no" or "no_overwrite".

README.md
Robinsane: Nice example!
I made a commit to use the pretty OS name instead of "Linux":
$(grep -oP '^PRETTY_NAME="\K[^"]+' /etc/os-release)

Dockerfile
Robinsane: I believe you misunderstand Docker's ARG:
These are env variables that only exist during the building process of the container image.
Once the image exists, these are gone forever.
Please refer the [dockerfile arg documentation](https://docs.docker.com/reference/dockerfile/#arg) for a better, more complete explanation.

ENV sets environment variables that remain, and can be used set as arguments with --env as desired when starting the container

Robinsane:
Please give some feedback / merge when you have spare time :)

Dockerfile
TheR1D: You are right, please ignore this comment.

sgpt/app.py
TheR1D: In order to keep the code base consistent, I would like to reuse existing patters as we checking the boolean values few lines above.

sgpt/role.py
TheR1D: Let's have config variables as `OS_NAME` and `SHELL_NAME` with default values `auto`.

README.md
TheR1D: Let's use `uname -s` as suggested above, since **/etc/os-release** does not exist on macOS. Given that we are supporting both Linux and macOS, ShellGPT should function correctly on both platforms.

Robinsane:
Changes made + tested in our production environment.
I believe this is ready to be merged :)

README.md
```diff
@@ -461,18 +461,36 @@ docker run --rm \
...
+* The \[E\]xecute option for --shell with interaction will not work, since it would try this Execute in the docker container.  
+=> setting the `SHELL_INTERACTION` environment variable to false , makes sense.
+* Since, most likely the os and shell of your container are not identical to the environment you want help with:  
+set the environment variables `OS_NAME` and `SHELL_NAME` according to your setup.
+
+
 Example of a conversation, using an alias and the `OPENAI_API_KEY` environment variable:
 ```shell
-alias sgpt="docker run --rm --env OPENAI_API_KEY --volume gpt-cache:/tmp/shell_gpt ghcr.io/ther1d/shell_gpt"
+alias sgpt="docker run --rm --volume gpt-cache:/tmp/shell_gpt --env OS_NAME=$(uname -s) --env SHELL_NAME=$(echo $SHELL) ghcr.io/ther1d/shell_gpt"
```
TheR1D, line 473:
```suggestion
alias sgpt="docker run --rm --volume gpt-cache:/tmp/shell_gpt --env OPENAI_API_KEY --env OS_NAME=$(uname -s) --env SHELL_NAME=$(echo $SHELL) ghcr.io/ther1d/shell_gpt"
```

Dockerfile
```diff
@@ -1,5 +1,9 @@
 FROM python:3-slim
 
+ENV SHELL_INTERACTION=false
```
TheR1D, line 3:
```suggestion
ENV SHELL_INTERACTION=false
ENV PRETTIFY_MARKDOWN=false
```

README.md
```diff
@@ -461,18 +461,36 @@ docker run --rm \
...
 ```
 
+Example environment variables for a working Ollama setup, using Docker:
+* ENV DEFAULT_MODEL=ollama/mistral:7b-instruct-v0.2-q4_K_M
+* ENV API_BASE_URL=http://10.10.10.10:11434
+* ENV USE_LITELLM=true
+* ENV OPENAI_API_KEY=bad_key
+* ENV SHELL_INTERACTION=false
+* ENV OS_NAME="Linux/Red Hat Enterprise Linux 8.8 (Ootpa)"
+* ENV SHELL_NAME=/bin/bash
```
TheR1D, line 494: This will not work with current Dockerfile since we do not install litellm during build. I think we should remove this for now.

TheR1D (changes requested):
Thank you for the updates @Robinsane. I ran some tests as well and have a few suggestions.

README.md
Robinsane: Great catch!

It had been awhile and I forgot my local Dockerfile looked different apart from the env variables:
`pip install /app[litellm]`

I adjusted and extended the documentation:
* made it alot more clear that the doc section is specifically for Docker + Ollama
* explained what needs to be done
* provided an inline Dockerfile example

TheR1D (approved):

TheR1D:
Hi @Robinsane, Sorry for the late reply. The PR has been merged. Thank you for your contribution!

PR #556, opened 2024-04-26
Fix #329: Ability to follow up on one-off prompts with chat and repl

### LINKED ISSUE #329: Feature request: Allow starting a chat from most recent response
I really love this interface for working with llms, but sometimes I intend to ask a one off question and realize the LLM almost but didn't quite get me, so I'd like to follow up rather than starting over with a new prompt.

I could use my shell's built in history to pipe in the last reply as context, but I think a more pleasant solution would be to introduce a `last` chat that stores the output of the last prompt that didn't specify a chat. Whenever a user makes a prompt without a `--chat` argument, the last entry is overwritten. If it's a named chat, have the `last` chat point to that named chat.

Since replies are cached by default, I think this would be really sensible and maybe there could even be a `--reply` flag that's syntactic sugar for responding to whatever was the last chat. To me, that's more ergonomic than assuming I'll want to continue the conversation before I ask my question.

jeanlucthumm:
Adds new flag `--follow-up` (`-fu`) which can be used with `--repl` and `--chat` to pull in the last one-off prompt.

Implemented by keeping a `last_chat` file in the cache directory that's overwritten for every default prompt call (`sgpt "[some prompt]"`), and then copying this file over to the chat cache directory when `-fu` is specified.

image

image

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -2,13 +2,13 @@ import os
 
 # To allow users to use arrow keys in the REPL.
 import readline  # noqa: F401
+import shutil
 import sys
+from pathlib import Path
 
 import typer
-import shutil
 from click import BadArgumentUsage
 from click.types import Choice
-from pathlib import Path
 
 from sgpt.config import cfg
 from sgpt.function import get_openai_schemas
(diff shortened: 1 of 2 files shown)


jeanlucthumm:
@TheR1D I thought there was a way for me to assign reviewers but I can't find it.

Robinsane:
> @TheR1D I thought there was a way for me to assign reviewers but I can't find it.

The button appears on the top right of this page, but only after it's been reviewed once already.
The idea is probably, once you made a PR, it's always ready for review :)

jeanlucthumm:
@TheR1D friendly ping on review

PR #557, opened 2024-05-02
"Raw" (e.g. "No-role") mode to simply call GPT as-is without a specific role or system message.

giladbarnea:
Added a -r, --raw option to use the LLM as-is without a specific role or system message.

Raw mode increases flexibility by enabling responses similar to those obtained directly through the LLM's API.

Discussion: #551

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,6 +436,7 @@ Possible options for `CODE_THEME`: https://pygments.org/styles/
 │ --interaction         --no-interaction    Interactive mode for --shell option. [default: interaction]    │
 │ --describe-shell  -d                      Describe a shell command.                                      │
 │ --code            -c                      Generate only code.                                            │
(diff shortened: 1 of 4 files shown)


jeanlucthumm:
You can create a custom role that's empty: 

```
{"name": "empty", "role": ""}
```

Unfortunately whether or not it uses markdown is dependent on the role description, so if you want that use this:

```
{"name": "empty", "role": "APPLY MARKDOWN"}
```

Then you can do

```
sgpt --role empty
```

giladbarnea:
@jeanlucthumm thanks! Your suggestion would do the job but feels to me more like a workaround, and specifying `--role empty` to me would feel less organic than `sgpt -r ...`. 
I acknowledge that it's a style thing, though.

PR #566, opened 2024-05-21
add possibility to modify generated shell command

odrevet:
Add the possibility to modify a generated completion in shell mode. 

Uses PromptSession package as [readline hook ](https://stackoverflow.com/a/8505387/2323803) do not works when stdin is set

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -7,6 +7,7 @@ import sys
 import typer
 from click import BadArgumentUsage
 from click.types import Choice
+from prompt_toolkit import PromptSession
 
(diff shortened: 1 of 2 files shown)


sgpt/app.py
```diff
@@ -7,6 +7,7 @@
 import typer
 from click import BadArgumentUsage
 from click.types import Choice
```
obeone, line 9: It's `prompt_toolkit` and not `PromptSession` ;)

obeone (approved):
It's work after correct this error

PR #575, opened 2024-06-04
Add Google Gemini support

lucasliet:
- add prompt to put google gemini api key on installation
- add condition to use gemini api key on litellm if model name starts with gemini/*
- change use litellm default value to true
- change default model to gemini/gemini-1.5-flash
---
since gemini has [free tier](https://ai.google.dev/pricing) usage and litellm has native support to it, it is interesting to use it instead of openai gpt


abhishekbhakat:
This PR #604 can also include Gemini. No need to create handling.

halida:
Any document about how to use it?

lucasliet:
> Any document about how to use it?

On first execution it will ask for your api key from gemini or you can expose it on env GEMINI_API_KEY, then you can select the gemini model like "gemini/gemini-1.5-flash" (I've set that one as default)

Isso the same usa usage you would have with openai, only renamed to gemini

halida:
Test OK:

DEFAULT_MODEL=gemini/gemini-1.5-flash
OPENAI_USE_FUNCTIONS=false
USE_LITELLM=true
OPENAI_API_KEY=TOKEN

lucasliet:
> This PR #604 can also include Gemini. No need to create handling.

I'm using lite llm but if you see the changes on handler.py it wasnt out of the box, also you have to handle the api key for gemini, and since its free, I make its usage default

abhishekbhakat:
> Can you tell how it would be configured with https://github.com/TheR1D/shell_gpt/pull/604 change?

This is what @halida explain above.

lucasliet:
plus, the way i've implemented, I could use both openai and gemini only passing --model flag, without changing any variable config

Issue #588, opened 2024-07-07
Add support for reading `OPENAI_API_KEY` from a file path specified in the config


giladbarnea:
Since my sgpt config, including my .sgptrc, is backed up via source control, I constantly need to manually make sure that the `OPENAI_API_KEY` line is kept out of the commit. 

This is cumbersome and only takes one lapse in concentration to accidentally push my key.

This issue could be resolved by allowing the `OPENAI_API_KEY` to reference a file path, where the file contains the key in plaintext.

![image](https://github.com/TheR1D/shell_gpt/assets/18548473/e4b3f856-fed5-488d-b983-64f3bebe8cb6)

PR #589, opened 2024-07-07
Add support for reading `OPENAI_API_KEY` from a file path specified in the config

giladbarnea:
My sgpt config, including the .sgptrc file, is backed up via source control. This means I constantly need to manually make sure the `OPENAI_API_KEY` line is kept out of commits.

This is cumbersome and only takes one lapse in concentration to accidentally push my key.

This PR provides a simple fix by letting OPENAI_API_KEY point to a file path. This separate file would hold the key in plaintext, making it easier to keep out of version control.

image

Corresponding issue: #588 

Changes:
- Added `POTENTIAL_PATH_KEYS = ("OPENAI_API_KEY",)` at the module level in sgpt/config.py. When reading the config, this determines if the value should be checked for referencing an existing file. If so, it reads the file's contents.
- The logic of reading a config line, including the file path support, was extracted to a method — `def _read_from_line(self, line: str) -> None`.
  - It adds a minor OOS change, skipping lines starting with "#" AFTER stripping them.
- Added a new tests/test_config.py with three tests to cover both regression and the new behavior.

Please let me know if you want me to remove, modify, add an integration test or anything else.
Thanks!

### DIFF
--- a/sgpt/config.py
+++ b/sgpt/config.py
@@ -37,6 +37,8 @@ DEFAULT_CONFIG = {
     # New features might add their own config variables here.
 }
 
+POTENTIAL_PATH_KEYS = ("OPENAI_API_KEY",)
+
 
 class Config(dict):  # type: ignore
     def __init__(self, config_path: Path, **defaults: Any):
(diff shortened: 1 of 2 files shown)


AlexanderYastrebov:
You can put dummy value (or comment out OPENAI_API_KEY) and put into your `.bashrc`:
```sh
export OPENAI_API_KEY=<your key>
```

PR #595, opened 2024-07-15
added optional user to additional kwargs for api calls

bmayfi3ld:
some models make use of the user field, and some organizations require it as a way to track usage.


bmayfi3ld:
looking more at this.  I think this might make the config value required?  I need to change it to optional.

PR #604, opened 2024-07-31
Support all LiteLLM  supported models

abhishekbhakat:
Allow `litellm` to handle API key retrieval independently. Requiring `shell-gpt` to explicitly pass API keys would necessitate adding logic for all models supported by         
 `litellm`. This approach would be:                                                                                                                                              
                                                                                                                                                                                 
 1. Redundant, as `litellm` already manages this functionality                                                                                                                   
 2. Potentially exhaustive, given the extensive list of models `litellm` supports                                                                                                
                                                                                                                                                                                 
 By letting `litellm` manage API keys, we avoid unnecessary code duplication and maintain a more streamlined integration.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies = [
 
 [project.optional-dependencies]
 litellm = [
-    "litellm == 1.24.5"
+    "litellm == 1.42.5"
 ]
 test = [
     "pytest >= 7.2.2, < 8.0.0",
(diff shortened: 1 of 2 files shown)


abhishekbhakat:
[Wiki](https://github.com/TheR1D/shell_gpt/wiki/Ollama#shellgpt-configuration) needs an update in case of supporting all LiteLLM Models. The instruction can be to just configure relative environment variables according to https://docs.litellm.ai/docs/providers

nictuku:
This seems convenient, because we could get all models supported right away without extra work by @TheR1D.

TheR1D:
merged, thank you @abhishekbhakat @nictuku

PR #621, opened 2024-08-30
Fix need of OpenAI API key if using Litellm

abhishekbhakat:
Fixes the need of unused OpenAI API key if Litellm is configured after #604.

### DIFF
--- a/sgpt/handlers/handler.py
+++ b/sgpt/handlers/handler.py
@@ -14,7 +14,7 @@ base_url = cfg.get("API_BASE_URL")
 use_litellm = cfg.get("USE_LITELLM") == "true"
 additional_kwargs = {
     "timeout": int(cfg.get("REQUEST_TIMEOUT")),
-    "api_key": cfg.get("OPENAI_API_KEY"),
+    "api_key": cfg.get("OPENAI_API_KEY") if not use_litellm else None,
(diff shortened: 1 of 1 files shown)


.gitignore
code-cntrl, line 1: Is this intended to be included in this pull request?

code-cntrl (changes requested):

.gitignore
abhishekbhakat: The project should have a gitignore. I found it helpful in development, so I kept it. Any specific reason you'd not want the gitignore?

abhishekbhakat:
@code-cntrl dropped the .gitignore
See if you can merge this.

I want the PR #575 to close but having to say you have to set a dummy key for `OPENAI_API_KEY` to use anything from LiteLLm makes it a friction.

lucasliet:
subject of the issue https://github.com/TheR1D/shell_gpt/issues/651

lucasliet:
in the case of #575, how can I provide the api_key from gemini in this implementation?

it would be good to have usage example on readme

abhishekbhakat:
@lucasliet you need to set `GEMINI_API_KEY` as env variable. As per https://docs.litellm.ai/docs/providers/gemini.

I explained in https://github.com/TheR1D/shell_gpt/pull/604#issuecomment-2259749221

PR #628, opened 2024-09-23
Add modify option before executing shell commands

eadadi:
# Motivation

Sometimes we need to make modifications to the suggested prompt before executing it.
This patch adds [M]odify option to the suggested prompt


eadadi:
is it going to be merged/checked/anything?

TheR1D:
Hello @eadadi, thank you for your contibution. It is a great feature, but It is a bit problematic to check new PRs that are not fully following [contribution](https://github.com/TheR1D/shell_gpt/blob/main/CONTRIBUTING.md) requirements. Could you please make sure the existing tests and linting tools are passing and add at least one more test in order to test new feature (if possible).

eadadi:
Ok, I have modified the tests to reflect the change.
In addition, I separated the modify to - modify comand [M] and modify and run command [R]. It is better to have both in my opinion.
And using only [M] allows further interaction with the AI - to description with respect to the modified command for example

eadadi:
Also, tests and lint are passing

eadadi:
@TheR1D

Issue #593, opened 2024-07-10
Ability to Update OpenAPI Key Without Manual Change


devilcoder01:
#### Current Behavior:

When initializing the code, users are prompted to enter their OpenAI API key. If the API key needs to be changed (e.g., after a month), it must be manually updated.

#### Proposed Improvement:
Implement a feature that allows users to easily update their OpenAI API key without manual intervention. This could be achieved through a configuration file or a command within the tool. This could be done through a command-line option.

KjellMorgenstern:
See also #549

devilcoder01:
Hi @KjellMorgenstern,

Currently, updating the OpenAI API key appears to be a manual process. I propose making this process automatic. For instance, when the program encounters an error in Shell GPT related to the API key, it could prompt the user with an option to update the OpenAI API key directly.

This would improve user experience and streamline the workflow.

Issue #549, opened 2024-04-20
I accidentally put the wrong api key, now I can't use it?


Pohie:
I put in the wrong api key, and now I can't change it... I can't find any options for doing so.

VincentTupper:
You can find your API key in the following file: (general file path for MacOS)
/Users/username/.config/shell_gpt/.sgptrc

There will be a line as such:
OPENAI_API_KEY=uegd97wagd9wbd8wahg9db

Also ensure hidden files are on -> Command + Shift + .

PR #658, opened 2024-12-24
Add OpenAI API Key Update Command

devilcoder01:
## Problem #593 #549 
Users occasionally enter incorrect OpenAI API keys during setup, and changing the key requires manual editing of the config file. This creates friction and could lead to user errors when modifying configuration files directly.

## Solution
Added a new command line option `--updateAPI` that allows users to update their OpenAI API key through a simple interactive prompt:
```sh
sgpt --updateAPI
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+myvenv
+__pycache__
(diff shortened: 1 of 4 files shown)


TheR1D:
Thank you @devilcoder01 for the contribution!

While I understand the intent behind simplifying the API key update process, I don't think a dedicated --updateAPI option is necessary at this time. It's fairly standard for applications to manage settings like API keys via a config or .rc file, and users modifying such files directly is a common and expected workflow.

Appreciate your effort, but I’ll close this for now.

Issue #659, opened 2024-12-31
UnicodeEncodeError with Emoji Output on Windows


LSauce:
## Description:
I'm encountering a UnicodeEncodeError when trying to generate a list of emojis . The error message indicates that the 'gbk' codec can't encode certain emoji characters.
## Steps to Reproduce:

1. Open the command line (PowerShell).
2. Execute the command:
```shell
sgpt emoji1
```
## Error Message:
```
UnicodeEncodeError: 'gbk' codec can't encode character '\U0001f60a' in position 64: illegal multibyte sequence
```
<details>
<summary>Error Detail</summary>

```
PS C:\Project\> sgpt emoji1
                                                                                                                                                                            
Here’s a list of some common emojis you might find useful:                                                                                                                                                                            
                                                                                                                                                                                                                                      
 • 😊 Smiling Face with Smiling Eyes                                                                                                                                                                                                  
 • 😂 Face with Tears of Joy                                                                                                                                                                                                          
 • ❤️ Red Heart                                                                                                                                                                                                                                                                                                                                                                                                                                                             
 • 👍 Thumbs Up                                                                                                                                                                                                                       
 • 🎉 Party Popper                                                                                                                                                                                                                    
 • 🐱 Cat Face                                                                                                                                                                                                                        
 • 🌞 Sun with Face                                                                                                                                                                                                                   
 • 🍕 Pizza                                                                                                                                                                                                                           
 • 🚀 Rocket                                                                                                                                                                                                                          
 • 🌈 Rainbow                                                                                                                                                                                                                         

If you have a specific emoji in mind or need more details, let me know! 😊
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ C:\Users\XXXX\AppData\Local\Programs\Python\Python311\Lib\site-packages\sgpt\app.py:229 in  │
│ main                                                                                             │
│                                                                                                  │
│   226 │   │   │   functions=function_schemas,                                                    │
│   227 │   │   )                                                                                  │
│   228 │   else:                                                                                  │
│ ❱ 229 │   │   full_completion = DefaultHandler(role_class, md).handle(                           │
│   230 │   │   │   prompt=prompt,                                                                 │
│   231 │   │   │   model=model,                                                                   │
│   232 │   │   │   temperature=temperature,                                                       │
│                                                                                                  │
│ ╭───────────────────────────────── locals ──────────────────────────────────╮                    │
│ │               cache = True                                                │                    │
[...]

PR #660, opened 2024-12-31
fix: resolve cache encoding issue by enforcing UTF-8

LSauce:
## Problem #659 
The error message indicates that the 'gbk' codec can't encode certain emoji characters.

## Solution
Set the cache to use UTF-8 encoding.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+# idea
+.idea
+
+# virtual env
+venv
--- a/sgpt/cache.py
+++ b/sgpt/cache.py
@@ -31,14 +31,14 @@ class Cache:
             key = md5(json.dumps((args[1:], kwargs)).encode("utf-8")).hexdigest()
             file = self.cache_path / key
             if kwargs.pop("caching") and file.exists():
-                yield file.read_text()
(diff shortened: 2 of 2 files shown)


TheR1D:
As I can see `path.write_text` and `path.read_text` is using [`open`](https://docs.python.org/3/library/functions.html#open) in order to read/write on FS. And encoding by default comes from OS, as it is stated in documentation:
> encoding is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding) returns), but any [text encoding](https://docs.python.org/3/glossary.html#term-text-encoding) supported by Python can be used. See the [codecs](https://docs.python.org/3/library/codecs.html#module-codecs) module for the list of supported encodings.

On my end, I don't have same issues with emojis as my default encoding on OS level is UTF-8. I believe forcing some specific encoding potentelly might create issues for other users.

Also I believe it possible to force specific encoding on python level:
`export PYTHONIOENCODING=UTF-8`

For now, I’ll go ahead and close this PR.

Issue #662, opened 2025-01-06
'NoneType' object has no attribute 'tool_calls'


arty-hlr:
Using open-webui as backend, this is my config file:
```
CHAT_CACHE_PATH=/tmp/chat_cache
CACHE_PATH=/tmp/cache
CHAT_CACHE_LENGTH=100
CACHE_LENGTH=100
REQUEST_TIMEOUT=60
DEFAULT_MODEL=llama3.1:latest
DEFAULT_COLOR=magenta
ROLE_STORAGE_PATH=/home/peng/.config/shell_gpt/roles
DEFAULT_EXECUTE_SHELL_CMD=false
DISABLE_STREAMING=false
CODE_THEME=dracula
OPENAI_FUNCTIONS_PATH=/home/peng/.config/shell_gpt/functions
OPENAI_USE_FUNCTIONS=false
SHOW_FUNCTIONS_OUTPUT=false
API_BASE_URL=http://XX.XX.XX.XX:3000/api
PRETTIFY_MARKDOWN=true
USE_LITELLM=false
SHELL_INTERACTION=true
OS_NAME=auto
SHELL_NAME=auto
OPENAI_API_KEY=sk-XXXXXXXXXXXXXXXX
```

Running a simple command I get this error:
![image](https://github.com/user-attachments/assets/5ede1d55-a125-416d-96df-7d9811879982)

It seems to come from this script:

https://github.com/TheR1D/shell_gpt/blob/aac2f5461b0d113dfdbfb6e926506c070d6afb52/sgpt/handlers/handler.py#L121

This sounds related and could be a fix as somehow open webui's responses don't have a delta:

https://github.com/run-llama/llama_index/issues/16570#issuecomment-2416893667

TheR1D:
Should be fixed
https://github.com/TheR1D/shell_gpt/pull/744

PR #663, opened 2025-01-06
fix for 'NoneType' object has no attribute 'tool_calls' issue

arty-hlr:
See https://github.com/TheR1D/shell_gpt/issues/662

Proposed fix by checking if `delta` is not None and if so just continuing the for loop. Another option would be to add a try catch around the `delta.get("tool_calls") if use_litellm else delta.tool_calls`. Let me know!

### DIFF
--- a/sgpt/handlers/handler.py
+++ b/sgpt/handlers/handler.py
@@ -116,6 +116,8 @@ class Handler:
             for chunk in response:
                 delta = chunk.choices[0].delta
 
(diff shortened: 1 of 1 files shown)


SimonTheCoder (approved):

SimonTheCoder (changes requested):
chunk.choices can be empty which will cause out of index when chunk.choices[0]

Issue #633, opened 2024-10-12
Error "Could not determine chat role" when sending too many messages to a single chat


igrmk:
I have an alias set up to use the predefined role using the `--chat` option.
For example, `--chat improve` is automatically added to every "Improve writing" query.

ShellGPT truncates the chat using the `CHAT_CACHE_LENGTH` option
each time it saves the conversation.

As a result, the first message eventually gets removed,
causing the tool to display "Could not determine chat role",
since the role is defined in that initial message.

To prevent this issue, the first message should likely never be truncated.

DanijelDomazet:
I've also aliased the `--chat` option, and I have the same issue. 

`Could not determine chat role of "dd" `

tinmarino:
Hello @igrmk, thank you for raising the issue, I experiment the same behavior.
For me, it seems to happen when I interrupt the chat with Ctrl-C and try to start it again.

![Screenshot from 2024-10-25 13-05-27](https://github.com/user-attachments/assets/d02fdb25-07f2-4d16-9bea-70a69dc8963a)

matthewblott:
Ah, I see I have the same problem :-)

EmVee381:
I'm not a Python developer, but purely through deductive reasoning, I think the problem is on this line.

https://github.com/TheR1D/shell_gpt/blob/aac2f5461b0d113dfdbfb6e926506c070d6afb52/sgpt/handlers/chat_handler.py#L74

And it is caused by how the cache is set up, as when the number of items from the end is taken, the first entry in the cache, which defines the initial role (e.g., system), is also deleted. Since the slice is taken from the end according to the number of items defined by the chat cache size, the role gets deleted. A possible solution could be either defining the role as "user" or fixing the code to something like:

```python
messages[1: -self.length]
```

EmVee381:
https://github.com/TheR1D/shell_gpt/pull/634#issuecomment-2658450268

matthewblott:
Thanks @EmVee381 👍 It doesn't look good that the previous PR to fix this submitted in October is still waiting approval 🙁

PR #669, opened 2025-02-14
Fix: Preserve the first item in the chat message cache containing the initial role

EmVee381:
I made a simple fix addressing the following issue. I'm not a Python developer, but it's just a straightforward modification.

related issue:
https://github.com/TheR1D/shell_gpt/issues/633


TheR1D (approved):
LGTM! Thank you @EmVee381

PR #678, opened 2025-03-25
Add copy to clipboard options for shell commands

a4v2d4:
- Enhancement: Automatically add shell commands to system clipboard on execution when using `sgpt --shell "<shell-cmd-prompt>"`

- Feature: Add "[C]opy" option to shell command prompt

_Both are disabled if ENV var 'COPY_SHELL_CMD_TO_CLIPBOARD' is 'false'. Defaults to true._

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -7,6 +7,7 @@ import sys
 import typer
 from click import BadArgumentUsage
 from click.types import Choice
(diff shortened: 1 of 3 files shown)


kth8:
I heard pyperclip is discontinued according to this project https://pypi.org/project/clipman

PR #680, opened 2025-03-31
feat: Add Gemini support and enhance shell completion (zsh only)

quique0194:
This PR introduces ZSH shell completion, improves the Typer CLI structure, and adds support for both OpenAI and Gemini LLMs.

Key changes:
• Adds support for Gemini LLM alongside OpenAI
• Implements ZSH completion support
• Modernizes Typer usage with a typer.Typer app instance
• Includes a .gitignore file to exclude python artifacts

Note: I'll pause work on this PR for now. Before merging, would like to implement:
• Add tests for the new ZSH completion feature
• Ensure compatibility between OpenAI and Gemini models
• Confirm OpenAI functionality is fully restored

In the meantime, any feedback is welcome!


TheR1D:
Thanks for the work on this, @quique0194,
I'm closing this PR for now because we don't want provider-specific Gemini handling in the codebase. ShellGPT already supports non-OpenAI models through the LiteLLM integration (`USE_LITELLM=true` in ~/.config/shell_gpt/.sgptrc). LiteLLM provides a single completion path for many providers, including Gemini, without maintaining separate handlers or provider-selection logic.

To use Gemini today:
```
USE_LITELLM=true
DEFAULT_MODEL=gemini/gemini-2.0-flash   # or another gemini/* model
```
Set the appropriate API key (e.g. GEMINI_API_KEY or GOOGLE_API_KEY, per [LiteLLM's Gemini docs](https://docs.litellm.ai/docs/providers/gemini)).

Adding a dedicated Gemini code path would duplicate that layer, increase maintenance (model validation, streaming, tool calls, etc.). We'd rather document and improve the LiteLLM path than merge parallel provider implementations.

PR #699, opened 2025-05-06
Add --fix option to correct the last command (#698)

### LINKED ISSUE #698: Is there an elegant way to provide feedback and rechat on the wrong commands and results?
I apologize for my unfamiliarity with this excellent project shell_gpt. An AI provided a command that failed
to execute. How can I share the incorrect context in my next conversation?

<logs>
~  sgpt --shell 'check which process using max memo'      
`ps -eo pid,comm,%mem | sort -nr | head -n 1`
[E]xecute, [D]escribe, [A]bort: e
zsh:1: command not found: 41219
</logs>

Then I want to tell ai that `The command failed with error: zsh:1: command not found: 41219, please retry.`.

Is there an elegant way to provide feedback and rechat?

I would be very grateful for your reply.

demouo:
Hi! Backgroud and motivation from #698

When I got an error command, I would like to correct ai that `the command ... was wrong with the output ..., please fix it`.

example follows:
image

This PR introduces a `--fix` flag. When used, sgpt retrieves the last executed command and its output, then prompts the AI to provide a corrected version based on the error.

Key changes:

- Added `--fix` option to `sgpt/app.py`.
- Created `sgpt/command.py` to manage storing and retrieving the last command and output in json format.
- Updated `sgpt/utils.py` to set and get the last command.
- Replaced `os.system` with `subprocess.Popen` in `run_command` to capture output.
- Added `get_fixed_prompt` to generate the prompt for the AI.
- Store command and output using the new Command class.
- Updated tests in `tests/test_shell.py` to reflect the use of `subprocess`.
- Added necessary configuration in `sgpt/config.py`.

make sure the lint and test result as usual:
image

Closes #698

### DIFF
--- a/sgpt/app.py
+++ b/sgpt/app.py
@@ -16,6 +16,7 @@ from sgpt.handlers.repl_handler import ReplHandler
 from sgpt.llm_functions.init_functions import install_functions as inst_funcs
 from sgpt.role import DefaultRoles, SystemRole
 from sgpt.utils import (
+    extract_command_from_completion,
     get_edited_prompt,
     get_fixed_prompt,
     get_sgpt_version,
@@ -226,6 +227,11 @@ def main(
             functions=function_schemas,
         )
 
+    if prompt == "":
+        raise BadArgumentUsage(
+            "Prompt cant be empty. Use `sgpt <prompt>` to get started."
+        )
+
     if chat:
         full_completion = ChatHandler(chat, role_class, md).handle(
             prompt=prompt,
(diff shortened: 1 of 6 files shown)


demouo:
The 2nd~4th Commits' changes:

1. LLM sometimes could not completely return the pure shell code, so I wrote the extracting function

2. Due to I was not familiar with the operation, I accidentally wrote several empty prompts, which should be ended before chatting with LLM

3. The idea of `fix last command` was added in REPL mode.

Hope everything goes well!

mtrajano:
@TheR1D Anything other than the conflicts in getting this merged? I'm interested in having this feature so would therefore like to take a stab at it. Looking forward to hearing back from you thanks!

Issue #692, opened 2025-04-21
Using o4-mini complains of bad "temperature" value in API


SohamG:
Wanted to upgrade to o4-mini, but apparently the API for o4-mini only accepts "1.0" for the temperature field. I didn't have much success in setting this on the terminal, and didn't find a well documented `.sgptrc` variable I could set.

maxregan:
As a workaround for now, you can indeed set `--temperature=1.0` on the command line, e.g

`sgpt "Say hello, world" --temperature 1.0 --model o4-mini`

varenc:
I ran into this same issue, but I'm using the `_sgpt_zsh` bindkey integration. I had to redefine it to be something like this. Would be nice if the `.sgptrc` config could set a default `TEMPERATURE` value so avoid this.  Or just ideally if it was aware of which models don't support temperature, but that's a bigger project.


```
function _sgpt_zsh() {
...
    local SGPT_EXTRAS=(--temperature=1.0 --no-functions)
...
    BUFFER="$(sgpt --shell $SGPT_EXTRAS <<< "$_sgpt_prev_cmd")"
...
```

dan-nicholls:
As an FYI, this also affects the new GPT5 models released today.

PR #720, opened 2025-08-09
feat: add `DEFAULT_TEMPERATURE` to config

dan-nicholls:
With the recent releases of the `gpt-5`, `gpt-5-mini` and `gpt-5-nano` models, along with the `o4-mini` models referenced in #692 , a custom temperature value isn't support and thus all require a temperate value of `1`. Currently the application throws an error if the default `0` value is used.

This PR introduces a `DEFAULT_TEMPERATURE` to the config so that if you already set a default model, you then don't need to use the `--temperature` flag every call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,6 +396,8 @@ CACHE_PATH=/tmp/shell_gpt/cache
 REQUEST_TIMEOUT=60
 # Default OpenAI model to use.
 DEFAULT_MODEL=gpt-4o
+# Default temperature value to use.
+DEFAULT_TEMPERATURE=0.0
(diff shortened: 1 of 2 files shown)


KoenLemmen:
Woudn't this also require 

```
temperature: float = typer.Option(
        cfg.get("DEFAULT_TEMPERATURE"),
        min=0.0,
        max=2.0,
        help="Randomness of generated output.",
    ),
```

In app.py

TheR1D:
Thanks for the PR, the issues was fixed in https://github.com/TheR1D/shell_gpt/pull/723 added you as co-author.

PR #723, opened 2025-09-08
Added default temperature to config

dokato:
The latest OpenAI models (e.g. GPT-5) require an explicit temperature value,  as they do not support omitting the parameter. This change sets a default  temperature in the config so users can work with these models seamlessly  without needing to specify it manually.


TheR1D:
@copilot resolve the merge conflicts in this pull request

TheR1D:
Hi @dokato thanks for this fix PR.

PR #730, opened 2025-10-28
Adapt to latest OpenAI API tool call format

huajijiaozhu:
Current format implemented in code is deprecated and not supported by certain OpenAI API-compatible model providers (e.g. DeepSeek), and causes crashes and/or malfunctions when tool call happens.

NOTE: Current code still lacks support for the rare case of multiple tool calls happen in a single completion response; should fix later if proved necessary

### DIFF
--- a/sgpt/handlers/handler.py
+++ b/sgpt/handlers/handler.py
@@ -1,4 +1,5 @@
 import json
+import uuid
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


TheR1D:
Migrated to OpenAI 2.0 in https://github.com/TheR1D/shell_gpt/pull/744

PR #733, opened 2025-10-29
Fix failing unit tests

TheR1D:


TheR1D:
TODO: some of the `assert result.exit_code == 0` were removed from shell tests.

PR #737, opened 2025-11-22
Fix handle empty stream chunks and enforce UTF-8 cache writes

printlndarling:
- Add safety check to skip empty/non-standard stream chunks from third-party APIs.
- Write cache files using UTF-8 to avoid Windows GBK errors caused by emoji output.

### DIFF
--- a/sgpt/cache.py
+++ b/sgpt/cache.py
@@ -38,7 +38,7 @@ class Cache:
                 result += i
                 yield i
             if "@FunctionCall" not in result:
-                file.write_text(result)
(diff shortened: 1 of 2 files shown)


TheR1D:
The issue was fixed in https://github.com/TheR1D/shell_gpt/pull/738

printlndarling:
@TheR1D Why did you close #737 but merge #738 with identical code?

TheR1D (approved):

TheR1D:
Hi @printlndarling, merged.

PR #738, opened 2025-12-08
Add check for empty choices in response chunks

1208nn:
This pull request introduces a small update to the `get_completion` function in `sgpt/handlers/handler.py` to improve robustness when handling streaming responses.

* Added a check to skip response chunks that do not contain any choices, preventing potential errors when accessing `chunk.choices[0]`.

### DIFF
--- a/sgpt/handlers/handler.py
+++ b/sgpt/handlers/handler.py
@@ -114,6 +114,8 @@ class Handler:
 
         try:
             for chunk in response:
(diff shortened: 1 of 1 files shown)


bbartels:
@TheR1D Could you have a look at this? This happens when some openai proxies automatically add `stream_options: { "include_usage": True }` and result in the last line containing the usage chunk without a choices object.

PR #753, opened 2026-03-17
feat: upgrade MiniMax default model to M3

octo-patch:
## Summary
Upgrade the existing MiniMax provider integration to use `MiniMax-M3` as the recommended default model while keeping `MiniMax-M2.7` / `MiniMax-M2.7-highspeed` as alternatives.

## Changes
- `README.md`: default `DEFAULT_MODEL` switched to `MiniMax-M3` (env, config file, and Docker example) and model lineup updated to list M3 first
- `tests/test_minimax.py`: default-model temperature clamping and chat-mode tests use `MiniMax-M3`; the previous-generation test now exercises `MiniMax-M2.7`
- `tests/test_minimax_integration.py`: real-API integration tests use `MiniMax-M3` (default prompt, code generation, temperature); the previous-generation case exercises `MiniMax-M2.7`
- Removed all references to `MiniMax-M2.5` / `MiniMax-M2.5-highspeed`

## Why
`MiniMax-M3` is the latest flagship: 512K context window, up to 128K output, and image input support. Promoting it as the default showcases the new model and matches the upgrade guidance, while keeping `MiniMax-M2.7` available for users who prefer the previous generation.

## Testing
- All 6 unit tests in `tests/test_minimax.py` pass
- Integration test class `TestMiniMaxIntegration` updated and structured identically (skips when `MINIMAX_API_KEY` is unset)

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,25 +486,25 @@ ENTRYPOINT ["sgpt"]
 
 
 ### Using MiniMax
-[MiniMax](https://www.minimaxi.com/) provides powerful LLM models (MiniMax-M2.5, MiniMax-M2.5-highspeed) with an OpenAI-compatible API. ShellGPT has first-class support for MiniMax -- just set the `MINIMAX_API_KEY` environment variable and the model name:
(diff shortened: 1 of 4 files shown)


TheR1D:
Thanks for the contribution. This PR adds MiniMax support and bumps the default model, but that isn’t something we’re looking to merge right now. We’re not planning to add new provider-specific integrations in this form, so I’m closing this. Appreciate the effort.

PR #772, opened 2026-06-05
fix: add click to dependencies

### LINKED ISSUE #771: sgpt fails to start: `click` is imported but not declared in dependencies
On a clean install, `sgpt` fails to start because `click` is imported by the package but is no longer present in the env:

```
$ sgpt --help
Traceback (most recent call last):
  File ".../bin/sgpt", line 4, in <module>
    from sgpt import cli
  File ".../site-packages/sgpt/__init__.py", line 1, in <module>
    from .app import main as main
  File ".../site-packages/sgpt/app.py", line 8, in <module>
    from click import UsageError
ModuleNotFoundError: No module named 'click'
```

`click` is imported directly in `sgpt/app.py`, `sgpt/utils.py`, `sgpt/config.py`, `sgpt/role.py` and `sgpt/handlers/chat_handler.py`, but it is not listed in `[project] dependencies` in `pyproject.toml`. It used to come in transitively through `typer`, but recent `typer` releases dropped that — `typer 0.26.7` (which `typer >=0.7.0, <1.0.0` resolves to today) only requires `shellingham`, `rich`, `annotated-doc`, and (Windows-only) `colorama`. So nothing pulls `click` in anymore.

This is independent of installer — any clean install that strictly follows the declared deps hits it. The minimal repro is just pip in a fresh venv:

```bash
python -m venv /tmp/sgpt
/tmp/sgpt/bin/pip install shell-gpt
/tmp/sgpt/bin/sgpt --help
# ModuleNotFoundError: No module named 'click'
```

`uv tool install shell-gpt` and `pipx install shell-gpt` reproduce it for the same reason (they always build an isolated env). Existing users likely didn't notice because `click` was already present in their environment from something unrelated. It also reproduces against `main` — the `pyproject.toml` and the `from click import ...` sites are unchanged.

### Environment

- shell-gpt 1.5.1
- typer 0.26.7
- Python 3.13
- macOS (arm64)

### Suggested fix

Either:

- add `click` to `[project] dependencies` in `pyproject.toml` (preferred, since `click` is used directly), or
- pin `typer` to a range that still pulls `click` transitively.

### Workaround

```bash
pip install click
# or, for isolated installers:
uv tool install --with click shell-gpt
pipx inject shell-gpt click
```

gavin913-lss:
## Problem

`click` is imported directly in `sgpt/app.py`, `sgpt/utils.py`, and `sgpt/config.py` but not declared in `pyproject.toml` dependencies. It used to come transitively through `typer`, but `typer 0.26.7` dropped that dependency.

Fixes #771

## Fix

Add `click >= 8.0.0` to `[project] dependencies` in `pyproject.toml`.

## Verification

```bash
pip install -e .
sgpt --help  # No longer fails with ModuleNotFoundError: No module named 'click'
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ classifiers = [
 dependencies = [
     "openai >= 2.0.0, < 3.0.0",
     "typer >= 0.7.0, < 1.0.0",
+    "click >= 8.0.0",
     "rich >= 13.1.0, < 14.0.0",
     "distro >= 1.8.0, < 2.0.0",
     'pyreadline3 >= 3.4.1, < 4.0.0; sys_platform == "win32"',


TheR1D (approved):
@gavin913-lss Thanks for the PR!

PR #773, opened 2026-06-06
fix: add click as explicit dependency

deuxksy:
## Problem

`sgpt` crashes with `ModuleNotFoundError: No module named "click"` on fresh install.

## Root Cause

`typer >= 0.26.0` vendored `click` internally and removed it from core dependencies:

```
typer 0.26.7 requires: shellingham, rich, annotated-doc, colorama (Windows only)
# click is no longer a transitive dependency
```

But `shell_gpt` directly imports from `click` in 5 files:
- `sgpt/app.py`: `UsageError`, `Choice`
- `sgpt/config.py`: `UsageError`
- `sgpt/role.py`: `UsageError`
- `sgpt/utils.py`: `BadParameter`, `UsageError`
- `sgpt/handlers/chat_handler.py`: `BadParameter`, `UsageError`

## Fix

Add `click >= 8.0.0, < 9.0.0` as an explicit dependency in `pyproject.toml` and regenerate `uv.lock`.

## Verification

- [x] `uv lock` regenerated with click 8.1.8 resolved
- [x] `uv run python -c "from sgpt import cli"` passes
- [x] Cross-validated with Codex and Gemini — fix confirmed correct


TheR1D:
Thanks for the PR, the click issue was fixed by https://github.com/TheR1D/shell_gpt/pull/772 added you as co-author.

PR #779, opened 2026-06-11
feat: make markdown_live vertical_overflow configurable

### LINKED ISSUE #778: Feature Request: Make `markdown_live` `vertical_overflow` Configurable
## Problem

Currently, `markdown_live` rendering uses Rich's `Live` display with `vertical_overflow="ellipsis"` behavior.

While this works well for most cases, it creates a poor experience during long-running REPL interactions or agent workflows.

When the output exceeds the terminal height, users only see:

```text

...

```

for extended periods of time.

This causes two issues:

1. **Lack of visibility**

   - Users cannot tell what the model is currently doing.

   - Long-running operations appear stalled even when progress is being made.

2. **Reduced educational value**

   - One of the benefits of AI-assisted workflows is observing the reasoning process, generated commands, tool calls, and intermediate outputs.

   - With `ellipsis`, much of this information is hidden from view.

## Proposed Solution

Expose `vertical_overflow` as a configurable option for markdown live rendering.

Example:

```python

Live(

    renderable,

    vertical_overflow=config.vertical_overflow,

)

```

Possible values:

```text

ellipsis

visible

crop

```

Default behavior should remain unchanged:

```text

vertical_overflow = "ellipsis"

```

so existing users are not affected.

## Benefits

- Preserves backward compatibility.

- Improves visibility during long-running REPL sessions.

- Makes debugging easier.

- Allows users to observe AI-generated intermediate output in real time.

- Better learning experience for users interested in understanding how the model is solving a problem.

## Additional Changes

I also plan to:

- Update `README.md`

- Document the new configuration option

- Provide examples of the different `vertical_overflow` modes and their effects

## Example Use Case

For interactive REPL sessions:

```python

vertical_overflow = "visible"

```

Users can continuously observe generated markdown output, tool execution details, and progress updates instead of staring at:

```text

...

```

for several minutes.

wangengzheng:
Add MARKDOWN_LIVE_VERTICAL_OVERFLOW config option to control how Rich's Live display handles content overflow in markdown rendering.

- New config: MARKDOWN_LIVE_VERTICAL_OVERFLOW (default: ellipsis)
- Supports values: ellipsis, visible, crop
- Backward compatible - default behavior unchanged
- Configurable via .sgptrc or environment variable

Closes #778

I wrote it through shell_gpt vibe coding


sgpt/printer.py
```diff
@@ -28,18 +30,48 @@ def __call__(self, chunks: Generator[str, None, None], live: bool = True) -> str
...
+                  live.update(
+                      Markdown(
+                          markup=full_completion,
+                          code_theme=self.theme,
+                      ),
+                      refresh=True,
+                  )
+                  last_refresh = time.monotonic()
+
+          # Avoid excessive terminal flickering caused by frequent refreshes.
```
TheR1D, line 63: I think this is not issue for most terminals to keep up refresh rate with coming tokens. Didn't see flickering in the terminals I use (Alacritty, iTerm, Gnome Terminal). But I believe some terminals might have this issues (depending or redering method). This feature should be optional in my opinion, can be updated later to use config var to enable.

README.md
```diff
@@ -391,9 +391,32 @@ SHOW_FUNCTIONS_OUTPUT=false
 OPENAI_USE_FUNCTIONS=true
 # Enforce LiteLLM usage (for local LLMs).
 USE_LITELLM=false
+# Control how markdown live rendering handles overflow when output exceeds terminal height.
+# Possible values: ellipsis, visible, crop
+MARKDOWN_LIVE_VERTICAL_OVERFLOW=ellipsis
```
TheR1D, line 396: I think it make sense to have `visible` as default, I will update it on relese commit.

TheR1D (approved):
Thanks for the PR!

README.md
wangengzheng: Sounds good. I agree that it makes sense to keep it visible by default. 
I’ve also opened a new pull request for the related changes: [[PR link](https://github.com/TheR1D/shell_gpt/pull/785/changes)]. Please take a look when you have a chance.

sgpt/printer.py
wangengzheng: Thanks for the feedback. That makes sense.
I’ve added a new configurable option for this feature, and it is disabled by default. Users who experience issues or prefer this behavior can enable it explicitly via the config.  
This should avoid affecting the default experience while still keeping the feature available when needed.
[[PR link](https://github.com/TheR1D/shell_gpt/pull/785/changes)]

PR #781, opened 2026-06-17
feat(config): add sgptrc extra_body support

### LINKED ISSUE #780: Add support for OpenAI SDK `extra_body` parameter in sgptrc
I'm using DeepSeek, as it supports disabling thinking via the `extra_body` parameter:

```python
    extra_body={
        "thinking": {"type": "disabled"}
    }
```

I prefer non-thinking mode to get a quick answer, and it is good enough for me. I believe some other users also need this parameter.

Can we add support for this parameter in sgptrc?

mcsrainbow:
## Summary

Adds support for configuring OpenAI SDK `extra_body` through `.sgptrc`.

This allows users of OpenAI-compatible providers, such as DeepSeek, to pass provider-specific request body options without code changes.

Example:

```text
EXTRA_BODY={"thinking":{"type":"disabled"}}
```

By default, EXTRA_BODY is empty and no extra request body is sent.

## Related Issue
Fixes https://github.com/TheR1D/shell_gpt/issues/780

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,8 +363,10 @@ You can setup some parameters in runtime configuration file `~/.config/shell_gpt
 OPENAI_API_KEY=your_api_key
 # Base URL of the backend server. If "default" URL will be resolved based on --model.
 API_BASE_URL=default
-# Extra JSON body to merge into OpenAI chat completion requests.
-EXTRA_BODY=null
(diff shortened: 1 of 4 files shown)


mcsrainbow:
Based on actual research, due to the differences among LLM Providers and the rapid changes in API standards, the requirement lacks sufficient support, therefore, this request is closed.

